PR #10, opened 2026-07-05
fix(llm): real tool-calling for the keyless local-agent backbone

### LINKED ISSUE #9: Keyless local-agent operators never run Arsenal tools — ReAct loop drops tool defs, abstains after one turn
## Summary

On the **keyless local-agent backbone** (Codex / Claude / Hermes — the headline "use the agent you already run" path), mission operators never execute a single Arsenal tool. The ReAct loop returns after **one** planning turn, so every operator abstains and recon/scan/exploit never touch the target. I pointed a mission at a host on my own LAN and watched all 10 operators finish with `{"findings":[],"abstained":true}` while the box sat there with 3 open ports.

It's deterministic, not a model mood or a sandbox-config thing. An API-key backbone (OpenRouter/Anthropic/OpenAI) works fine on the same mission. The bug is specific to the local-agent path.

## Repro

1. Connect Codex in the War Room (or start a mission with `provider: 'local-agent', model: 'codex'`).
2. Point it at any reachable target and launch.
3. Watch the operators. They reason, read the repo, and abstain. `nmap`/`curl`/etc never run against the target. `GET /api/mission/status` shows tasks completed with 0 tool-backed findings.

## Root cause

The ReAct loop drives *every* Arsenal action through `chatWithTools(...).toolCalls`:

```ts
// src/agent/index.ts:137
const response = await this.llm.chatWithTools(messages, toolDefs, { ... });
if (response.toolCalls?.length) {
  // ... execute tools via arsenal.execute ...
} else {
  // src/agent/index.ts:214 — treats the first text turn as the FINAL answer and returns
}
```

But `LocalAgentAdapter.chat()` (the local-agent backbone) ignores `options.tools` entirely and never returns `toolCalls`:

```ts
// src/llm/index.ts (LocalAgentAdapter)
private formatPrompt(messages, options) {
  const parts = [ /* preamble */ ];
  if (options?.maxTokens) parts.push(...);
  for (const m of messages) parts.push(`\n### ${m.role.toUpperCase()}\n${m.content}`);
  return parts.join('\n');            // options.tools is dropped on the floor
}
async chat(messages, options) {
  const content = await localAgentChat(agentId, prompt, ...);
  return { content: content.trim(), model: ..., finishReason: 'stop' };  // no toolCalls, ever
}
```

So `response.toolCalls` is always `undefined`, the loop takes the `else` on iteration 0, and it concludes **without ever calling `arsenal.execute`**. The CLI agent just gets a plain text prompt with no idea the Arsenal exists.

The read-only Codex sandbox is a red herring — the harness never asks Codex to *run* anything. It asks for a text plan and expects `tool_calls` a text-only adapter can't produce.

## Why this is a bug, not the intended design

The wiring clearly means to execute; only the adapter can't deliver it:

- Every operator gets `operator.attachArsenal(this.arsenal, agentLoop)` regardless of backbone (`src/index.ts:1140`).
- `AgentLoop`'s own header: *"ReAct (Reason + Act) agent that connects LLM tool calling to Arsenal execution."*
- The no-tool text path in the operator is explicitly labeled `// Fallback: LLM-only execution (no tool calling)` (`src/operators/index.ts:469`) — a fallback, not the plan.

## Impact

The keyless path is the whole pitch — "no API keys, driven by the AI coding agent you already run." On that path the swarm can only reason, never act. You also pay for it: I saw **13 Codex sessions spawned in ~30 min**, one per operator tick, each one burning a full session to read the repo and then abstain.

## Suggested fix

Give the CLI-agent adapter a text tool-calling protocol: inject the tool catalog into the prompt, ask the agent to reply with a fenced ` ```tool ` JSON block, and parse that back into `toolCalls`. Execution still runs in-process via `arsenal.execute`, so the CLI's read-only sandbox stays irrelevant — the agent only *picks* the tool, the harness runs it.

I've prototyped this against `LocalAgentAdapter` (left `CodexAdapter`, the `codex` provider, alone since its prompt is deliberately planning-only). With the catalog exposed, real Codex emits a valid tool call in ~22s:

```
finishReason: tool_calls
toolCalls: [ { name: "port_scan", arguments: { target: "192.168.68.157", timeout: 1000 } } ]
```

vs the old path where the same task got a text abstention and zero tools. Full suite still 297/297.

Happy to open a PR if useful.

DMontgomery40:
Fixes #9.

## What's broken

On the **keyless local-agent backbone** (Codex / Claude / Hermes) the ReAct loop never runs a single Arsenal tool. `LocalAgentAdapter.chat()` drops `options.tools` and never returns `toolCalls`, so `AgentLoop` takes its no-tool branch on iteration 0 (`src/agent/index.ts:214`) and concludes without ever calling `arsenal.execute`. Every operator reasons, then abstains. I watched a 10-operator mission finish `{"findings":[],"abstained":true}` against a live host with 3 open ports.

This is the headline path — "keyless warfare, no API keys." The README says the mission engine is stable keyless (status table) and that every operator "runs the same real, tool-backed ReAct loop as recon." It doesn't, keyless. The benchmarked numbers all ran on API-key backbones (gpt-5.5, Opus 4.8, single-agent) — never the local-agent path where this bug lives. An API-key backbone works fine on the same mission; the defect is specific to local-agent.

## The fix

Give the CLI-agent adapter a text tool-calling channel: put the tool catalog + a reply protocol in the prompt, and parse a fenced ` ```tool ` JSON block back into structured `toolCalls`. Execution stays in-process via `arsenal.execute`, so the CLI agent's read-only sandbox is irrelevant — it only picks the tool; the harness runs it. Multi-turn history is serialized so an ephemeral `codex exec` still sees prior calls and results.

Scoped to `LocalAgentAdapter`. Left `CodexAdapter` (the `codex` provider) alone — its prompt is deliberately planning-only ("Do not run active probes").

## Receipts

- New test `src/__tests__/local-agent-tools.test.ts` (5 cases): catalog exposure, tool-block parsing, prose-is-final, no-misread of a findings JSON, unchanged no-tools path.
- `npm test` → 297/297. `npm run typecheck` clean. `npm run doctor` ok. `npm run field:drill` passed.
- `npm run verify-claims` → **24/24, unchanged** (this touches the adapter, not any bench artifact).
- Real-Codex e2e: with the catalog exposed, Codex emits a valid call in ~22s where the old path abstained —
  ```
  finishReason: tool_calls
  toolCalls: [ { name: "port_scan", arguments: { target: "…", timeout: 1000 } } ]
  ```

## Honest scope

This makes the keyless swarm *able* to execute tools, and Codex followed the protocol cleanly in a single-call e2e. What I did **not** do is benchmark the full multi-turn, 8-archetype swarm end-to-end on the keyless path — protocol adherence across a whole mission is a prompt-tuning question I only spot-checked. So: plumbing fixed and unit+e2e-proven; keyless-swarm reliability is the next thing to earn a number.

### DIFF
--- a/src/__tests__/local-agent-tools.test.ts
+++ b/src/__tests__/local-agent-tools.test.ts
@@ -1,0 +1,76 @@
+/**
+ * Text-mode tool calling for the KEYLESS local-agent backbone (Codex / Claude / Hermes).
+ *
+ * Regression for the "operators never execute a single Arsenal tool" defect: the ReAct
+ * AgentLoop drives execution through `chatWithTools(...).toolCalls`, but the CLI-agent
+ * adapters returned text only and silently dropped `options.tools`, so `toolCalls` was
+ * ALWAYS undefined → the loop concluded on iteration 0 and no tool ever fired.
+ *
+ * The fix teaches the CLI-agent adapters a text protocol: describe the tools in the
+ * prompt, and parse a fenced ```tool JSON block back into structured `toolCalls`.
+ * Execution still runs in-process via `arsenal.execute` — the CLI's own sandbox is
+ * irrelevant because we only ask it to PICK the tool, not run it.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const { mockLocalAgentChat } = vi.hoisted(() => ({ mockLocalAgentChat: vi.fn() }));
+vi.mock('../agent/local-agents.js', () => ({ localAgentChat: mockLocalAgentChat }));
+
+import { LLMBackbone } from '../llm/index.js';
+import type { LLMToolDefinition, LLMMessage, LLMConfig } from '../types/index.js';
+
+const portScan: LLMToolDefinition = {
+  name: 'port_scan',
+  description: 'Scan TCP ports on a target host',
+  parameters: { type: 'object', properties: { target: { type: 'string' } }, required: ['target'] },
+};
+const messages: LLMMessage[] = [{ role: 'user', content: 'Assess 192.168.68.157' }];
+const cfg = (): LLMConfig => ({ provider: 'local-agent', model: 'codex', maxTokens: 4096, temperature: 0.4 });
+
+describe('local-agent backbone — text-mode tool calling', () => {
+  beforeEach(() => mockLocalAgentChat.mockReset());
+
+  it('exposes the tool catalog to the agent (prompt names the callable tools)', async () => {
+    mockLocalAgentChat.mockResolvedValue('done, no tools needed');
+    const llm = new LLMBackbone(cfg());
+    await llm.chatWithTools(messages, [portScan]);
+    const promptSent = String(mockLocalAgentChat.mock.calls[0][1]);
+    expect(promptSent).toContain('port_scan');
+    expect(promptSent).toContain('Scan TCP ports');
+  });
+
+  it('parses a fenced ```tool JSON block into a structured toolCall the ReAct loop can execute', async () => {
+    mockLocalAgentChat.mockResolvedValue(
+      'I will map the reachable services first.\n\n```tool\n{"tool":"port_scan","arguments":{"target":"192.168.68.157"}}\n```'
+    );
+    const llm = new LLMBackbone(cfg());
+    const res = await llm.chatWithTools(messages, [portScan]);
+    expect(res.toolCalls).toBeDefined();
+    expect(res.toolCalls).toHaveLength(1);
+    expect(res.toolCalls![0]).toMatchObject({ name: 'port_scan', arguments: { target: '192.168.68.157' } });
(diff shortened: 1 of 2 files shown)


jmagly:
Your diagnosis here was spot-on — but heads up that this looks superseded. The maintainer landed an equivalent text tool-calling fix directly on `main` in commit `921ccc2` ("fix(keyless): text-CLI tool-calling (hardened per audit)"), which gives `LocalAgentAdapter` the same tool-contract injection + `parseTextToolCalls` path this PR proposes — the current adapter on `main` already returns `finishReason:'tool_calls'` with structured `toolCalls`, and there's a dedicated `local-agent-tool-calling` test. This PR now shows conflicts against that reworked `src/llm/index.ts`. So #9 is effectively resolved and this can probably be closed as landed-differently — worth a maintainer confirming.

jmagly:
be sure to rebase with main once you have completed this pr

PR #16, opened 2026-07-05
fix(keyless): tool-calling on the local-agent path + 2 review fixes

younger-plinius:
Follow-up to #8 — fixes the one bug that breaks the **headline keyless path**, plus two review nits.

### 🔴 Keyless path never executed the Arsenal
On the `local-agent`/`codex` backbones (the "hunt with the agent you already run" path), `LocalAgentAdapter`/`CodexAdapter` dropped `options.tools` and returned **no `toolCalls`**. So the ReAct loop (`src/agent/index.ts:214`) took its "final answer" branch on turn 0 and **every operator abstained** — recon/scan/exploit never touched the target — while API-key backbones worked fine on the same mission. (Repro: mission at a LAN host → all operators finish `{"findings":[],"abstained":true}` with open ports sitting there.)

**Fix:** teach the text-CLI adapters to do **tool-calling over text** — render the Arsenal + a strict JSON action-contract into the prompt, then parse the agent's reply back into `LLMToolCall[]` (`parseTextToolCalls`). Backward-compatible: only activates when tools are offered; a prose debrief is still treated as the final answer. New regression test pins the backbone surfacing `toolCalls` so this can't silently come back.

### Review fixes
- `scripts/cve-zero-split.mjs`: self-test asserted `holdout.length === 0` but it's now the curated 10 → validates the 10-sample split (`--self-test` back to 6/6).
- `docs/index.html`: static CVE-Zero copy said `6/10 + 9/10 held-out` → now matches the README (`single-agent 8/10 exact, 10/10 found`).

296 tests green, `verify-claims` 24/24, `tsc` clean.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -5551,7 +5551,7 @@
                         <p style="opacity: .65; font-size: 13px; margin: 0 0 14px;">A domain lights up only when there's a benchmark behind it. Greyed tiles are in development.</p>
                         <div class="feature-grid" id="domainGrid">
                             <div class="feature-card"><div class="feature-icon">🕸️</div><h3>Web ✅</h3><p>Apps, APIs, auth flows, OWASP Top 10. XBEN 90.1% pass@1.</p></div>
-                            <div class="feature-card"><div class="feature-icon">📂</div><h3>Code ✅</h3><p>White-box source audits. CVE-Zero 6/10 + 9/10 held-out (7 languages).</p></div>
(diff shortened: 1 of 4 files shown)


younger-plinius:
Superseded by #19 — identical code, reopened with clean, neutral commit authorship (the original commit here carried a local machine identity) plus a README readability pass. Please review/merge #19 instead. 🫡

PR #18, opened 2026-07-05
fix(local-agents): Windows CLI detection + Hermes auth path

psigho:
## What

Fixes local-agent detection on Windows. Two bugs meant a correctly-installed keyless agent was reported as unusable in the War Room:

1. **npm-global CLIs (Claude Code, Codex) reported as `binary not found on PATH`.** `resolvePath()` used POSIX-only `command -v` / `which`, and `execFile`/`spawn('claude', ...)` can't resolve npm's `.cmd` shims via `PATHEXT`, so detection got `ENOENT` and marked the agent `installed: false`.
2. **An authed Hermes desktop reported as `NOT AUTHED`.** Hermes on Windows stores its login under `%LOCALAPPDATA%\hermes`, not `~/.hermes/`, so the presence check missed it.

## How

- Add `resolveBin()` — uses `where.exe` on Windows, preferring `.exe > .cmd > .bat` — and route detection plus both spawn call sites through the resolved path.
- Add `localAppData()` and check both `%LOCALAPPDATA%\hermes` and `~/.hermes/` (presence-only).
- Launch Windows `.cmd` shims through a new `spawnAgent()` helper that invokes `cmd.exe` with an **argv array (`shell: false`)** instead of `shell: true`, so the untrusted prompt is never re-parsed by `cmd.exe` — no command-injection surface.

## Reviewer notes

- No new dependencies; `npm run typecheck` clean.
- Touches only `src/agent/local-agents.ts` (+ a `.gitignore` entry).
- POSIX detection paths are unchanged — the new `resolveBin()` branches on platform.
- Security-relevant: the `shell: false` argv approach is the reason the shim launch doesn't reintroduce injection; worth a look during review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ bench/nyu/
 # sensitive export leak-gate denylist (operator handle + embargoed disclosure
 # targets + personal vault) loaded by the exporter — never commit
 scripts/.export-denylist.local
+server.pid
+server.log
--- a/src/agent/local-agents.ts
+++ b/src/agent/local-agents.ts
@@ -54,6 +54,16 @@ export function agentHome(): string {
 const expand = (p: string): string => (p.startsWith('~') ? agentHome() + p.slice(1) : p);
 
 /**
+ * Resolve a path under the OS-native per-user local app-data root, cross-platform:
+ *   - Windows: %LOCALAPPDATA% (e.g. C:\Users\<u>\AppData\Local), where the Hermes desktop app
+ *     stores its runtime — its .env / auth lives at %LOCALAPPDATA%\hermes\.env, NOT ~/.hermes/.env.
+ *   - POSIX:   <agentHome>/AppData/Local fallback (harmless: the artifact simply won't exist there).
+ * PRESENCE-ONLY: callers pass the result to existsSync(); contents are never read.
+ */
+const localAppData = (rel: string): string =>
+  join(process.env.LOCALAPPDATA || join(agentHome(), 'AppData', 'Local'), ...rel.split('/'));
+
+/**
  * Env for a spawned agent CLI. Two adjustments to our own env:
  *   - strip the injected provider keys so the CLI falls back to its OWN native login (not t3mp3st's).
  *   - point HOME (and USERPROFILE on Windows) at the agentHome() so the CLI finds that login even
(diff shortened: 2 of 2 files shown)


jmagly:
Useful fix — the Windows npm-shim detection was genuinely broken. Two things worth hardening before merge:

**1. `cmd.exe /c` prompt passing is an injection surface.** `spawn('cmd.exe', ['/d','/s','/c', bin, ...args])` — cmd re-parses its tail with its own rules and doesn't honor `\"`-escaping, so a prompt containing `"` followed by cmd metacharacters (`&`, `|`, `>`, `%VAR%`) can break out and execute (the BatBadBut / CVE-2024-1874 class). Prompts reach `spawnAgent` straight from argv (`claude -p <prompt>`, `codex exec <prompt>`) and are adversarial in this tool. Only the `.cmd`/`.bat` shims (claude/codex) on Windows are affected — `.exe`/POSIX/Hermes are fine. This is new code (Windows never worked before), so it's incomplete hardening rather than a regression, but worth addressing: reject/escape prompts containing `"`/`%`/cmd metacharacters, or route through a safe exec wrapper.

**2. Narrow POSIX detection regression.** Detection now gates on `resolveBin` returning a value, so on a bash-less minimal image where both `command -v` and `which` fail but the binary is on PATH, a genuinely installed agent flips to "not installed." Mainstream macOS/Linux are unaffected (the `which` fallback covers Alpine/busybox). Cheap fix: on POSIX, fall back to a direct `execFile(spec.bin)` attempt before concluding not-installed.

The `%LOCALAPPDATA%\hermes\` auth-path fix and the detection approach are good keepers.

jmagly:
Please rebase this branch onto current `main`, then address the Windows shell/Node-version safety issue before review continues.

The Windows detection fix is directionally useful, but the branch launches `.cmd`/`.bat` shims with shell behavior while `package.json` still permits older Node 18 versions. The code comment relies on safer argument escaping in patched Node releases, but the project engine floor does not enforce that.

Suggested agent repair path:
1. Rebase from current `main`.
2. Either raise the Node engine floor to a patched version that safely handles this path, or avoid sending prompt/user-controlled arguments through `.cmd`/`.bat` shell shims.
3. Pay special attention to Hermes `-z <prompt>` because that prompt is passed as an argument rather than stdin.
4. Add Windows-focused tests for `.cmd` resolution and prompt argument escaping.
5. Verify the diff stays limited to local-agent detection/launch behavior and package metadata/tests needed for that fix.

jmagly:
Approved

jmagly:
Local verification passed on head `bf50be9` as `jmagly`:

```bash
npm ci
npm run typecheck
npm test
```

jmagly:
Delivered via #99. The #18 head commit `bf50be9` is now an ancestor of `upstream/main`, with the conflict repair preserving current-main POSIX well-known path resolution while adding the Windows `where.exe` / safe `.cmd` launch / Hermes `%LOCALAPPDATA%` auth behavior. GitHub shows this original PR as closed rather than merged, but the code landed in `main` through the batch merge.

PR #20, opened 2026-07-05
feat: local models + Grok Build (xai) backbones run the Arsenal

younger-plinius:
Builds on the keyless tool-calling that just merged in #19 — extends the same **bring-your-own-brain backbone** so **local models and Grok Build** also run the Arsenal, plus a README pass. Three commits, one theme.

### 🖥️ Self-hosted local models now run the Arsenal
The `local` provider existed but **ignored `options.tools`** — so a local model hit the same keyless-path abstain bug the CLI agents did (no `toolCalls` → ReAct bails turn 0 → Arsenal never runs). Now `LocalAdapter` does **tool-calling over text** (reuses `renderToolContract`/`parseTextToolCalls` from #19), so the Arsenal runs on **any** local model — even ones without native function-calling. Speaks **both Ollama `/api/chat` and any OpenAI-compatible `/v1`** server (LM Studio / vLLM / llama.cpp), env-configurable (`TEMPEST_LOCAL_BASE_URL` / `TEMPEST_LOCAL_MODEL`), and selectable in the provider menu.

### 🤖 Grok Build as a first-class `xai` provider
`grok-build-0.1` via xAI's **OpenAI-compatible API** (`api.x.ai/v1`, `XAI_API_KEY`, native tool-calling). Deliberately **not** wired as a headless CLI backbone: the `grok` CLI has no plan-only/read-only mode (only `--always-approve`), so driven as a planning brain it would auto-execute edits/shell **ungated** and double-agent the ReAct loop. The API path gives the model as a clean, gated reasoning brain.

### 📖 README
Refreshed the hook (*"hands it an arsenal"* + fully-offline), documented the local-model + Grok Build backbones, added a navigation TOC.

---
**Verified in-branch:** `tsc` clean · **309 tests** (local-model + xai cases included) · `verify-claims` **24/24**. Commit authorship neutral.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,11 @@
 
 </div>
 
-Point T3MP3ST at an authorized target and the kill chain is yours — **recon → exploit → report**, from a web War Room or the CLI, driven by the AI coding agent you're already signed into (Claude Code, Codex, Hermes). No new API keys, no cloud, no second bill. Your agent is the brain; T3MP3ST is the war machine you bolt around it. Self-hosted storm, keyless warfare. ⚡
+**Your AI coding agent is already a hacker — T3MP3ST hands it an arsenal.**
 
-The recon engine is live and tool-backed; the exploit loop is benchmark-proven — **90.1% pass@1 on XBEN**, XBOW's own 104-challenge suite, plus hint-free CTF solves and a cold hunt on real post-cutoff CVEs. Every number recomputes from committed data (`npm run verify-claims`), and the [status table](#what-ships-today) is exact about what's live, what's scaffolding, and what's still roadmap. The full 8-operator swarm is the architecture it grows into — loud about the mission, honest about where the build is. Full receipts, caveats, and the held-out CVE breakdown are in [Benchmarks](#benchmarks) below.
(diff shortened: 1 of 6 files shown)


younger-plinius:
**Addressed the review — rebased + receipt.**

**P1 (merge-safety / #13):** rebased onto current `main`. The `config/index.ts` conflict was resolved by keeping #13's safe behavior — **no `process.cwd()/.env` read, no persisting env keys into Conf** — and adding `XAI_API_KEY` to the safe `envVarMap` lookup path in `getApiKey`. CI green in-branch: typecheck clean, **323 tests**, **0 lint errors**, build clean, `verify-claims` **24/24**.

**P2 (xAI model receipt):** verified `grok-build-0.1` against xAI's published model list ([docs.x.ai/docs/models](https://docs.x.ai/docs/models), 2026-07-05):
- `grok-build-0.1` — coding model, **256K** ctx (the default)
- `grok-4.3` — general, **1M** ctx (added as a second option)

Both are the exact `model` strings `/v1/chat/completions` accepts; banked as a cited comment in `config/index.ts`. This is a docs/model-list receipt, not a live authenticated call (I don't hold an `XAI_API_KEY`) — the final confirmation on your end is `curl -sH "Authorization: Bearer $XAI_API_KEY" https://api.x.ai/v1/models | grep grok-build-0.1`.

PR #23, opened 2026-07-05
fix: avoid loading caller cwd dotenv in server

mane:
## Summary
- Remove the API server's top-level `dotenv.config()` call so it does not implicitly import `.env` from the caller's working directory.
- Add a static regression check alongside the existing API-key env hardening tests.

## Bug verified
`ConfigManager` already avoids `process.cwd()/.env`, but `src/server.ts` imported `dotenv` and called `dotenv.config()` before the server initialized. When T3MP3ST is run from or inside an audited target repo, that re-enables cwd `.env` loading and can contaminate the server process with unrelated target secrets.

RED before fix:
- `npx vitest run src/__tests__/api-key-env-static.test.ts` failed on the new invariant because `src/server.ts` still contained `import dotenv from 'dotenv'` / `dotenv.config()`.

## Test plan
- `npx vitest run src/__tests__/api-key-env-static.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint` (existing warnings only, 0 errors)
- `npm run verify-claims`

### DIFF
--- a/src/__tests__/api-key-env-static.test.ts
+++ b/src/__tests__/api-key-env-static.test.ts
@@ -3,6 +3,7 @@ import { readFileSync } from 'fs';
 import { join } from 'path';
 
 const configSource = readFileSync(join(process.cwd(), 'src/config/index.ts'), 'utf8');
+const serverSource = readFileSync(join(process.cwd(), 'src/server.ts'), 'utf8');
 const setupScript = readFileSync(join(process.cwd(), 'scripts/setup-api.sh'), 'utf8');
 
 function configLoadEnvBlock(): string {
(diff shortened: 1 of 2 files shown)


jmagly:
This is correct and complete for its goal — removing the top-level `dotenv.config()` fully stops the caller-cwd `.env` leak, and legitimate env loading is unaffected since `ConfigManager` (imported before the old dotenv call) already loads only `~/.t3mp3st/.env` / `~/.env`. It's just currently conflicting on `src/__tests__/api-key-env-static.test.ts`, which #55 also edits — a rebase resolves it. Recommend landing this one first (it's smaller and isolated), then rebasing #55's test edits on top.

jmagly:
Please rebase with main.

jmagly:
Superseded by #96, which reapplies this fix on current main and preserves the regression test. Thank you @mane for identifying the caller-CWD dotenv contamination and providing the original patch.

PR #28, opened 2026-07-05
fix: reject curl destination override flags

mane:
## Summary
- Reject curl transport/DNS/proxy override flags in `/api/tools/execute` parsing.
- Add a regression invariant for flags such as `--resolve`, `--connect-to`, proxy flags, Unix socket flags, and interface binding.

## Bug verified
`/api/tools/execute` authorized network commands against the parsed URL hostname, but curl can be given flags such as `--resolve approved.test:443:127.0.0.1` that keep the visible URL host as `approved.test` while connecting to a different effective destination. That makes the approval target misleading.

RED before fix:
- `npx vitest run src/__tests__/local-api-hardening-static.test.ts` failed because no curl destination-override rejection existed.

## Test plan
- `npx vitest run src/__tests__/local-api-hardening-static.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint` (existing warnings only, 0 errors)
- `npm run verify-claims`

### DIFF
--- a/src/__tests__/local-api-hardening-static.test.ts
+++ b/src/__tests__/local-api-hardening-static.test.ts
@@ -37,10 +37,25 @@ describe('local API authorization hardening invariants', () => {
   it('/api/tools/execute rejects curl flags that change the effective destination', () => {
     const parser = sourceBlock('const CURL_TRANSPORT_OVERRIDE_FLAGS', 'function inferCommandTarget');
 
-    for (const flag of ['--resolve', '--connect-to', '--proxy', '--preproxy', '--socks5', '--socks5-hostname', '--unix-socket', '--interface']) {
+    for (const flag of [
+      '--resolve',
+      '--connect-to',
+      '--proxy',
+      '--preproxy',
+      '--socks5',
+      '--socks5-hostname',
+      '--unix-socket',
+      '--interface',
+      '--url',
+      '--config',
+      '--next',
+      '-K',
+    ]) {
       expect(parser).toContain(`'${flag}'`);
     }
     expect(parser).toMatch(/findCurlTransportOverrideFlag\(args\)/);
+    expect(parser).toMatch(/countCurlUrlOperands\(args\)\s*>\s*1/);
+    expect(parser).toMatch(/multiple URL operands/);
     expect(parser).toMatch(/changes the effective network destination/);
   });
 
--- a/src/__tests__/local-api-hardening-static.test.ts
+++ b/src/__tests__/local-api-hardening-static.test.ts
@@ -4,12 +4,16 @@ import { join } from 'path';
 
 const serverSource = readFileSync(join(process.cwd(), 'src/server.ts'), 'utf8');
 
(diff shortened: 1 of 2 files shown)


hummbl-dev:
Good hardening direction. One remaining egress-scope gap I would close in this same guard or a follow-up: `/api/tools/execute` still infers the approved network target from the last non-flag argument, while curl can contact destinations supplied through other URL/config forms.

Concrete cases to consider rejecting for curl:

- multiple URL operands in one command, because the last operand can match the approved target while an earlier operand is still fetched;
- `--url` / `--url=...`, because the effective URL is no longer just the last positional argument;
- `--config` / `-K`, because a local curl config can reintroduce `url`, `proxy`, `resolve`, or `connect-to` outside this argv scan;
- `--next`, because it permits multiple transfer segments with different options/targets.

The current tests prove the listed override flags are blocked, but adding one static or unit invariant for these forms would make the target-authorization boundary harder to bypass.

mane:
Thanks, good catch — I closed those remaining curl forms in `d170d92`.

What changed:
- reject `--url` / `--url=...`
- reject `--config` / `-K` (including compact `-Kfile`)
- reject `--next`
- reject curl commands with more than one URL operand, so an earlier fetched URL cannot differ from the final approval-matching URL

I also extended the local hardening invariant to cover these cases.

Verification after the update:
- `npx vitest run src/__tests__/local-api-hardening-static.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint` (existing warnings only, 0 errors)
- `npm run verify-claims`
- GitHub Actions `test`: pass

jmagly:
Good direction, but the rejection isn't complete — a couple of real destination-override bypasses get through:

**1. Bundled short options.** `findCurlTransportOverrideFlag` matches exact tokens / `-x*` / `-K*`, but curl allows bundling boolean shorts before a value-taking short: `-sx <proxy>` = `-s -x <proxy>`. The token `-sx` isn't detected, and a bare `host:port` proxy value dodges the `countCurlUrlOperands > 1` backstop (it isn't URL-shaped), so `curl -sx internal:8080 http://approved.com` authorizes the approved URL while actually connecting through the attacker proxy. `-sK evil.cfg` (a curl config file can set `url=`/`proxy=`) is the same class.

**2. Missing flags.** `--proxy1.0` isn't in the denylist, and the DNS-override family (`--dns-servers`, `--doh-url`, `--dns-*`) is only caught incidentally by the URL-operand heuristic — not reliably.

**3. Test can't catch the above.** It's source-static (asserts the flag strings appear in the source), so it wouldn't detect the bundling bypass. A behavioral test that feeds these strings through `parseCommand` and asserts rejection would.

Suggest: normalize bundled shorts (expand `-sx` → `-s -x`, or reject any short cluster containing `x`/`K`), add `--proxy1.0` + the `--dns-*`/`--doh-url` family, and add a behavioral test. It's a real improvement as-is, just shouldn't be considered complete hardening yet.

PR #30, opened 2026-07-06
feat: cloud / mobile / binary-RE offensive loadouts (arsenal + detection benchmarks)

younger-plinius:
Three new offensive-security category loadouts, each = **arsenal + an offline-scorable benchmark**. All opt-in and honestly scoped — the roster marks every one **🚧** (detection, not full exploitation), and dangerous frameworks stay `catalog_only` (never minted as callable tools).

### ☁️ Cloud
Base CLIs (`aws`/`az`/`gcloud`) + attack-path/assessment (`scoutsuite`/`cloudfox`/`pmapper`) on top of the existing `prowler`/`checkov`; **`pacu`** (AWS exploitation) is `catalog_only`. New **`cloud:bench`** — an IaC-misconfig detection benchmark (Terraform corpus + checkov scorer): offline `--self-test`, `cloud:bench:live` for the real number with checkov installed.

### 📱 Mobile
`apkleaks`/`mobsfscan` (static, safe) + `drozer`/`objection` (gated) + `class-dump` (iOS static) on top of `apktool`/`jadx`; **`frida`** (runtime injection) is `catalog_only`. New **`mobile:bench`** — a built-in static ruleset (manifest misconfig + hardcoded-secret / cleartext detection) over a seeded corpus.

### 🔩 Binary / RE
`ghidra`/`objdump` + `checksec`/`strings`/`readelf` triage on top of `radare2`/`binwalk`; **`gdb`** (executes the target) is `receipt_required`. New **`binary:bench`** — a decompiled-output sink detector (unsafe-copy / format-string / cmd-injection / int-overflow) over a seeded corpus.

### Honesty notes
- The **mobile/binary benches are DETECTOR REGRESSION CHECKS** — the rules and the corpus fixtures are authored together, so a pass is a floor that guards the detector from breaking, **not a real-world detection rate**. Labeled as such in the output + corpus notes.
- **No new headline numbers** — the README roster carries no percentages; `verify-claims` stays **24/24**.
- Risk/execution gating throughout; `pacu`/`frida` unit-tested non-mintable.

**Verified:** tsc clean · **337 tests** · `verify-claims` 24/24 · lint 0 errors · cloud/mobile/binary benches green.


dasouzakenley-hue:
From UI, i can't specify for RE.
There is not RE in Guided.

how can I re binary via T3MP3ST?

PR #37, opened 2026-07-06
feat: safe upstream update script (npm run update)

Wibias:
## Summary
- Add cross-platform updater (`update.mjs` → `update.ps1` / `update.sh`)
- Back up and restore paths in `scripts/update-protected.txt` (secrets, bench output, disclosures)
- Add `npm run update`, `update:dry`, `update:hard`
- Document protected paths and outside-repo config in README

## Test plan
- [ ] `npm run update:dry -- --force` shows plan without changing files
- [ ] Fresh copy / no-commit repo: confirm first-time sync + `npm install`
- [ ] With `.env` or `bench/wild-hunt/` present: confirm backup + restore
- [ ] Windows PowerShell + Linux/macOS bash paths


jmagly:
Well-architected updater. One real data-loss gap plus a couple of cross-platform divergences to fix:

**1. First-time (no-commits) path deletes `reports/` and `evidence/`.** On the tarball/copied-tree install this feature targets, both drivers `rm -rf` the working tree except `.git` + protected paths — and `scripts/update-protected.txt` omits `reports/` and `evidence/`, so a single `y` permanently deletes engagement findings/artifacts (the same dirs #47 mounts as persistent volumes). Add both (and any other output dirs) to the protected manifest. A `git stash`/checkout approach for the first-time path would be safer than `rm -rf`.

**2. `update:dry` isn't read-only on Unix.** `update.sh` runs `git init` / `remote add` / `git fetch` unconditionally, before the dry-run check — so `npm run update:dry` on a tarball install mutates git state, contradicting the docs and diverging from `update.ps1` (which guards these). Move them behind the dry-run guard.

**3. `!`-negation ignored on Unix.** `update.sh collect_protected` skips `!` patterns, so `!.env.example` isn't honored and the template gets backed up + restored over upstream's version. `update.ps1` handles negation correctly.

**4. No failure trap on Unix.** `update.ps1` prints the backup location on failure; `update.sh` (`set -euo pipefail`, no trap) just dies mid-run — after a wipe, the user isn't told where the temp backup is. Add a bash trap that reports it.

The destructive modes being behind a `y/N` prompt is fine; the gap is specifically the unprotected output dirs on the wipe path.

Wibias:
Thanks for the feedback @jmagly 

1. output dirs on first-time wipe
reports/ and evidence/ are in update-protected.txt now (and documented in the README). They get backed up and restored on the no-commits path instead of being deleted with the rest of the tree.

2. update:dry mutating git on Unix
git init / remote add / fetch only run on the real sync path now, not in dry-run. Dry-run just prints what would run. Matches update.ps1.

3. ! negation on Unix
update.sh now does the same two-step as PowerShell: collect positive patterns, then apply negations. !.env.example actually works, we don't back up or restore over upstream's template anymore.

4. no failure trap on Unix
Added an ERR trap that prints the temp backup path if the script dies mid-run. Cleared on success and on cancel at the prompt.

Wibias:
@elder-plinius Please implement an update script. I would say thats the most important to not have struggle with updating..

jmagly:
Please rebase this branch onto current `main` and resubmit the updater work without unrelated deletions.

The branch currently deletes benchmark corpora, scripts, provenance docs, and tests that are outside the safe-upstream-update scope. That blocks merge. The updater also needs isolated review because it can change local repository state.

Suggested agent repair path:
1. Start from current `main`.
2. Keep only updater scripts, docs, and focused updater tests.
3. Restore unrelated benchmark/provenance/test files.
4. Add or document tests for dry-run, dirty worktree behavior, protected paths, and any hard-reset/destructive mode.
5. Make destructive behavior opt-in and clearly documented.
6. Confirm `git diff --name-status upstream/main...HEAD` contains no unrelated deletions.

Wibias:
@jmagly Done!

jmagly (changes requested):
Re-reviewed current head `5f12b03`. The earlier scope/rebase problem is fixed: the branch is now limited to updater scripts, updater docs, package scripts, and focused updater tests. It no longer carries the unrelated benchmark/provenance/test deletions.

Verification I ran:
- `npm ci` -> pass
- `npm run typecheck` -> pass
- `npm run test:update` -> pass locally; PowerShell-specific cases skipped because `powershell` is not installed in this environment

One remaining clean-delivery blocker before merge:

```bash
git diff --check upstream/main...HEAD
# scripts/test-update.mjs:290: new blank line at EOF.
```

Please remove the extra blank line at EOF in `scripts/test-update.mjs` and push again. After that, this should be ready from my side; the substantive updater fixes I asked for are present (`reports/` and `evidence/` protected, dry-run guarded before git mutation, bash negation support, bash failure trap, and hard reset remains opt-in).

jmagly (changes requested):
Re-audited updated head `4f2d75d`.

Pre-audit threat assessment: `flag` for caution because this PR adds an updater that backs up/restores local paths and runs git/npm operations. I did not see agent-instruction poisoning, CI/workflow tampering, dependency changes, remote script execution beyond the intended git fetch, generated payloads, or secret exfiltration attempts. This is safe to continue reviewing, but the updater safety boundary needs direct smoke coverage.

Verification I ran:

```bash
npm ci
npm run typecheck
npm run test:update
npm test
npm run lint        # pass, existing warnings only
git diff --check upstream/main...HEAD
```

Those passed. However, a direct dry-run smoke check of the shipped entrypoint still fails in a clean worktree with no protected paths:

```bash
npm run update:dry
# exit=1
# worktree remained clean
```

Relevant output:

```text
scripts/update.sh: line 239: [[: 0
0: syntax error in expression (error token is "0")
scripts/update.sh: line 268: [[: 0
0: syntax error in expression (error token is "0")
scripts/update.sh: line 286: [[: 0
0: syntax error in expression (error token is "0")
Proceed with update? [y/N]
[x] Update failed partway through.
```

Root cause appears to be `protected_count="$(grep -c . "$protected_list" 2>/dev/null || echo 0)"`: when the list is empty, `grep -c` prints `0` and exits nonzero, then `echo 0` also runs, so `protected_count` becomes `0\n0` and breaks integer comparisons.

Please fix the zero-protected-path count path and add/adjust a bash test that exercises `npm run update:dry` or `scripts/update.sh --dry-run` without `--force` / without protected files. The shipped dry-run preview should exit cleanly and should not trip the failure trap just because stdin declines the prompt or the protected list is empty.

Still requesting changes until that direct dry-run path is clean.

Wibias:
Fixed the empty protected-list count bug in update.sh (grep -c + || echo 0 was producing 0\n0). Added bash tests for --dry-run with no protected files, both with and without --force. npm run test:update passes (13/13). Pushed to feat/safe-upstream-update (04162ab).

jmagly (changes requested):
Current-head delivery audit at `04162ab` found one remaining cross-platform blocker using a checksum-verified PowerShell 7.6.3 runtime.

`PATH=/tmp/powershell-7.6.3:$PATH npm run test:update` runs the previously skipped PowerShell integration suite, and all four PowerShell dry-run cases fail with:

```text
update.ps1: Cannot bind argument to parameter Path because it is null.
```

The failure is at `Join-Path $env:TEMP ...`: PowerShell Core on Unix does not guarantee `TEMP` is set. The Bash cases all pass. Please derive the backup root from `[System.IO.Path]::GetTempPath()` (portable on Windows and PowerShell Core) and add a static assertion or always-on test that pins this choice. I am preparing a current-main maintainer replacement so this does not leave the contribution stalled.

jmagly:
Delivered via #98. The replacement preserves @Wibias as author of the scoped updater feature, adds the portable PowerShell temp-path correction found during final delivery testing, passes both Bash and checksum-verified PowerShell 7.6.3 integration suites, and passed the repository’s full CI gate. Closing this branch as superseded; thank you for the updater and the iterative fixes.

PR #46, opened 2026-07-06
fix(llm): support versioned OpenAI-compatible endpoints + optional bearer for local provider

alameya:
## Problem

The `local` provider couldn't talk to two common kinds of OpenAI-compatible server:

1. **Versioned endpoints other than `/v1`.** `LocalAdapter.isOpenAIWire()` matched only a literal `/v1` path, so versioned endpoints like **Zhipu/z.ai's `/api/paas/v4`** silently fell through to Ollama's native `/api/chat` wire format and failed.
2. **Servers that require a bearer.** The `local` provider was treated as always-keyless, so authenticated OpenAI-compatible servers (Zhipu/z.ai, Together, …) could never send a real `Authorization` header — and `tempest test` rejected a configured local provider with "No API key configured".

## Fix

- `isOpenAIWire` now matches any versioned path (`/v1`, `/v4`, …) via `/\/v\d+(\/|$)/` instead of a literal `/v1`.
- The `local` provider reads an optional bearer from **`TEMPEST_LOCAL_API_KEY`** (with `ZAI_API_KEY` / `ZHIPUAI_API_KEY` as aliases), wired through `getApiKey`/`hasApiKey` so `tempest test`, `status`, and the mission engine all accept a configured local provider.
- The key is read from env **in-memory only** and never persisted into the store — matching the existing env-key hardening invariant (`api-key-env-static`).
- The env template now documents `TEMPEST_LOCAL_BASE_URL` / `TEMPEST_LOCAL_MODEL` / `TEMPEST_LOCAL_API_KEY`, including guidance that any `/vN` path speaks the OpenAI wire.

## Tests

- New `src/__tests__/local-openai-wire-static.test.ts` (5 static invariants, repo style) pins both fixes.
- Existing review standard stays green:

```
npm run typecheck   ✅
npm test            ✅ 347/347 (32 files)
npm run doctor      ✅
```

## Verified end-to-end

Configured against z.ai's global API and confirmed a full round-trip:

```
TEMPEST_LOCAL_BASE_URL=https://api.z.ai/api/coding/paas/v4
TEMPEST_LOCAL_MODEL=glm-5.2
$ tempest test
✔ Connection successful!
Response: Connection successful!
```

Also confirmed against `glm-4.5-air`, `glm-4.6`, `glm-5.1`. Backward compatible — Ollama-on-`/api` and LM-Studio/vLLM-on-`/v1` paths behave exactly as before; the bearer is only sent when a key is present.

## Scope

No behavior change for existing keyless local setups. The versioned-path regex is a strict superset of the old literal-`/v1` match.

### DIFF
--- a/src/__tests__/local-openai-wire-static.test.ts
+++ b/src/__tests__/local-openai-wire-static.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it } from 'vitest';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+// Static invariants for the `local` provider. Two real OpenAI-compatible servers
+// broke against the old code: (1) the wire detector only matched a literal /v1, so
+// versioned paths like Zhipu/z.ai's /api/paas/v4 silently fell through to the
+// Ollama native format and failed; (2) the provider was treated as always-keyless,
+// so servers that require a bearer (Zhipu, Together, …) could never authenticate.
+// These tests pin the fixes in place and document the intent.
+
+const llmSource = readFileSync(join(process.cwd(), 'src/llm/index.ts'), 'utf8');
+const configSource = readFileSync(join(process.cwd(), 'src/config/index.ts'), 'utf8');
+
+function block(source: string, startMarker: string, endMarker: string): string {
+  const start = source.indexOf(startMarker);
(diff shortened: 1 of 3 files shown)


TheoOliveira:
Got in here today, Glad to see to enable a custom api

PR #63, opened 2026-07-09
fix(ui): reset War Room to idle when a hunt is declined or aborted

psigho:
## What
A declined-approval hunt (**"hunt not started"**) — or hitting the red-X abort — left the War Room stuck advertising an active run:
- the **ENGAGE** button stayed on the `ENGAGING…` spinner and disabled,
- the **LIVE** fidelity banner kept claiming *"real tools probing the target"*,
- the **narration rail** kept showing *"Looking at your target — finding the attack surface"*.

Nothing was actually running (the hunt never started), so this was a purely cosmetic-but-alarming stuck state.

## Why
`completeMission()` only toggled the button's `display`, never restoring its label / `disabled` / `opacity`, and never cleared the fidelity banner or narration rail. `abortMission()` had the same gap **and** no-op'd entirely when `missionRunning` was already `false` — so the red X did nothing after a declined hunt.

## Fix
Add `resetEngageUI()` — restores the ENGAGE button (`▶ ENGAGE`, enabled) and clears the fidelity banner (`setFidelity('off')`) + narration rail. Route both `completeMission()` and `abortMission()` through it, and let the red X clear a leftover engaging/probing UI even when `missionRunning` is already `false`.

Client-side only (`docs/index.html`); no behavior change to an actually-running mission. Verified by simulating the stuck state and confirming both the decline and abort paths reset the button, banner, and rail.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -7591,6 +7591,26 @@ o.id === 'coordinator' ? `1. MISSION_STATUS: Progress assessment
         // (each already paired with an addIntel into System Events) stay harmless.
         function addMissionLog(_type, _message) { /* no-op: Mission Log panel retired */ }
 
+        // Return the War Room to an idle, ready-to-engage state. Every mission-terminal path
+        // (complete, fail, approval-declined, abort) MUST call this: it restores the ENGAGE button
+        // (a declined hunt otherwise leaves it stuck on the "ENGAGING…" spinner) and clears the
+        // LIVE/SIM fidelity banner + the "what it's doing now" narration rail, so the UI can never
+        // keep advertising an active probe after the hunt has stopped.
+        function resetEngageUI() {
+            const cmdStart = document.getElementById('cmdStartBtn');
+            if (cmdStart) {
+                cmdStart.style.display = 'block';
+                cmdStart.disabled = false;
+                cmdStart.style.opacity = '1';
+                cmdStart.innerHTML = '▶ ENGAGE';
+            }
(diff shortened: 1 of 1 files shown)


jmagly:
Looks good — this fixes a real stuck-button bug. `startMissionFromDashboard` sets ENGAGE to a disabled `ENGAGING…` spinner with no `finally`, so the decline path (`completeMission('approval declined')`) left it stuck; `resetEngageUI()` restores it and correctly covers both the decline and abort paths without touching timers or `missionRunning`. One cosmetic nit: the new abort-idle branch sets `cmdMissionName='STANDBY'` but leaves `cmdMissionStatus` unchanged, so the header can read STANDBY with a stale red "approval declined" substatus — worth clearing that too. Non-blocking.

jmagly (approved):
Re-reviewed current head `d43c405`. The follow-up commit clears the stale idle substatus in the red-X/no-running-mission path, so the previous nit is addressed.

This patch is scoped to the War Room UI reset paths and does the right thing: `completeMission()` restores the ENGAGE button and clears the fidelity/narration indicators, while `abortMission()` can now clean up leftover UI even when `missionRunning` is already false after a declined hunt. I also extracted the inline scripts from `docs/index.html` and ran `node --check`; no syntax errors found.

Approved.

jmagly:
Local verification passed on head `d43c405` as `jmagly`:

```bash
npm ci
npm run typecheck
npm test
```

PR #64, opened 2026-07-09
fix(ui): route OBSIDIVM benchmarks through any backend, not OpenRouter-only

psigho:
## What
The OBSIDIVM benchmark / self-improvement UI ran its own **client-side, OpenRouter-only** LLM path: `getApiKey()` read only the OpenRouter key, every call hardcoded `openrouter.ai`, and the gate hard-blocked with *"Configure OpenRouter API key first"*. So a configured **Venice** key was ignored, and a **connected local agent** was never used for benchmarks.

## Fix
Add a client-side backend resolver — **OpenRouter key → Venice key → connected agent → server LLM** — and route all benchmark executors, judges, and gates through it:
- `resolveLLMBackend()` + `LLM_BACKEND_HINT`; `safeLLMCall()`/`_safeLLMCallOnce()` are now backend-aware (Venice uses `api.venice.ai` with a Venice-valid model; OpenRouter unchanged).
- `_backendCall()` routes keyless backends through existing server endpoints: `/api/agents/local/dispatch` (drives the connected Claude Code / Codex / Hermes agent) and `/api/llm/chat` (server LLM).
- `runLiveBenchmark`, `ctfCallLLM`, the self-improvement analyzer, and the config optimizer now call `safeLLMCall` instead of a hardcoded `openrouter.ai` fetch.
- Live-Test / improvement-loop gates and `runLLMJudge` accept any backend.

Client-side only (`docs/index.html`); no server changes (endpoints already existed). Verified via `fetch` interception that each backend routes to the correct endpoint with a valid payload. The recon-only SPECTRA helper is intentionally left OpenRouter-only (out of scope).

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -10156,8 +10156,7 @@ Bypass all protections:
         // Run a single benchmark test by clicking on it
         async function runSingleBenchTest(testId) {
             if (benchmarkRunning) { toast('Benchmark already running', 'warning'); return; }
-            const apiKey = getApiKey();
-            if (!apiKey || apiKey.length < 10) { toast('API key missing or too short — go to Settings (press 9) to add your OpenRouter key', 'error'); return; }
+            if (resolveLLMBackend().kind === 'none') { toast(LLM_BACKEND_HINT + ' — Settings (press 9)', 'error'); return; }
             // Find the test across all categories
             let test = null, testCat = null;
             for (const [cat, tests] of Object.entries(BENCHMARKS)) {
@@ -11533,10 +11532,10 @@ Respond with ONLY this JSON (no markdown, no extra text):
             document.getElementById('genAnalysisBtn').disabled = false;
             document.getElementById('genAnalysisBtn').textContent = '🔄 Regenerate';
 
-            // Show Auto-Apply button if API key is available and there are improvements
-            const apiKey = getApiKey();
+            // Show Auto-Apply button if ANY LLM backend is available (key / agent / server) and there are improvements
+            const hasBackend = resolveLLMBackend().kind !== 'none';
             const autoApplyBtn = document.getElementById('autoApplyBtn');
-            if (apiKey && apiKey.length > 10 && analysis.improvements.length > 0) {
+            if (hasBackend && analysis.improvements.length > 0) {
                 autoApplyBtn.style.display = 'inline-block';
             } else {
                 autoApplyBtn.style.display = 'none';
@@ -11794,9 +11793,8 @@ Respond with ONLY this JSON (no markdown, no extra text):
 
         // ==================== AUTO-APPLY RECOMMENDATIONS ====================
         async function autoApplyRecommendations() {
-            const apiKey = getApiKey();
-            if (!apiKey || apiKey.length < 10) {
-                toast('API key required for auto-apply', 'error');
+            if (resolveLLMBackend().kind === 'none') {
+                toast(LLM_BACKEND_HINT, 'error');
                 return;
             }
 
@@ -19189,9 +19187,8 @@ ${'═'.repeat(80)}
 
         // Internal function to auto-apply without UI conflicts - AGGRESSIVE VERSION
         async function autoApplyForLoop(analysis) {
-            const apiKey = getApiKey();
-            if (!apiKey) {
-                console.log('[IMPROVE] No API key, using fallback improvements');
+            if (resolveLLMBackend().kind === 'none') {
+                console.log('[IMPROVE] No LLM backend, using fallback improvements');
                 return applyFallbackImprovements(analysis);
             }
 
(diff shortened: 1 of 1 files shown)


jmagly:
Good refactor — collapsing the duplicated OpenRouter fetch blocks into `safeLLMCall` + `resolveLLMBackend` is a real improvement, and the OpenRouter and Venice paths check out. One bug to fix before the new routing works in the main deployment:

The keyless backends fetch **relative** URLs — `fetch('/api/agents/local/dispatch', …)` and `fetch('/api/llm/chat', …)` — but the rest of the app builds requests as `${T3MP3ST_API.baseUrl}${endpoint}` (the existing server-LLM call goes through `T3MP3ST_API.post`; agent status uses `base + '/api/agents/local/status'`). `baseUrl` is set to `http://<host>:3333` (or a custom URL) whenever the static War Room (GitHub Pages `docs/`) talks to a separate backend — the primary cross-origin setup. There, those two relative fetches hit the Pages origin instead of the backend → 404/CORS, so the local-agent and server-LLM routing (the headline new capability) silently fails; it only works when the page is served same-origin as the backend. Fix: prefix `${T3MP3ST_API.baseUrl}` on both (or route `/api/llm/chat` through `T3MP3ST_API.post`).

Two smaller things: `/api/agents/local/dispatch` isn't referenced anywhere else in the client (only `status`/`connect`/`detect`/`ping`) — worth confirming the server route exists; and the bundled `claude-opus-4.6 → 4.8` default bump is now inconsistent with the state default and `renderModels`/`saveApiKey` (still `4.6`).

jmagly (changes requested):
Re-reviewed the current head `8e73492`. The prior cross-origin issue is fixed: `_backendCall()` now prefixes `T3MP3ST_API.baseUrl`, and the `4.8` default drift in the touched paths is back to `4.6`. The modified inline scripts also pass a syntax-level `node --check` after extracting them from `docs/index.html`.

Still requesting changes because two OpenRouter-only gates remain in flows this PR explicitly claims now support Venice, connected local agents, and server LLM.

1. **Click-to-run single benchmark still hard-requires an OpenRouter key** (`docs/index.html:10157-10160`). The main `runBenchmarks('live')` path now checks `resolveLLMBackend()`, but clicking an individual `.bench-test[data-test]` still does:

```js
const apiKey = getApiKey();
if (!apiKey || apiKey.length < 10) {
  toast('API key missing or too short — go to Settings (press 9) to add your OpenRouter key', 'error');
  return;
}
```

So with only a Venice key, connected local agent, or server LLM, the headline single-test benchmark path is blocked before it reaches `runLiveBenchmark()` / `safeLLMCall()`. Replace that gate with the same backend check used by `runBenchmarks`:

```js
if (resolveLLMBackend().kind === 'none') {
  toast(LLM_BACKEND_HINT + ' — Settings (press 9)', 'error');
  return;
}
```

2. **Self-improvement auto-apply still falls back when there is no OpenRouter key** (`docs/index.html:19191-19196`, plus the manual auto-apply UI gate at `11536-11543` / `11796-11801`). `autoApplyForLoop()` now uses `safeLLMCall()` for the optimizer call, but it returns `applyFallbackImprovements(analysis)` before that if `getApiKey()` is empty. That means the improvement loop still does not use Venice, connected agent, or server LLM for its optimizer step. The manual auto-apply button is also hidden/blocked by an OpenRouter-only key check even though `autoApplyRecommendations()` now calls `safeLLMCall()`.

Patch shape:

```js
const hasBackend = resolveLLMBackend().kind !== 'none';
```

Use that instead of `apiKey && apiKey.length > 10` for showing/enabling auto-apply, and in both auto-apply entry points:

```js
if (resolveLLMBackend().kind === 'none') {
  toast(LLM_BACKEND_HINT, 'error'); // or fallback in the loop if desired
  return;
}
```

After those changes, verify the four supported modes against these UI paths:

- OpenRouter key: full benchmark, single benchmark click, judge, config optimizer, self-improvement auto-apply.
- Venice key only: same paths, no OpenRouter key present.
- Connected local agent only: same paths route to `${T3MP3ST_API.baseUrl}/api/agents/local/dispatch`.
- Server LLM only: same paths route to `${T3MP3ST_API.baseUrl}/api/llm/chat`.

I did not find new syntax errors in the changed HTML script.

jmagly (approved):
Re-reviewed current head `2f1fbd8`. The previously requested OpenRouter-only gates are fixed:

- single benchmark click now uses `resolveLLMBackend().kind === 'none'` instead of requiring `getApiKey()`;
- manual auto-apply visibility and execution now use backend availability;
- self-improvement `autoApplyForLoop()` now only falls back when no backend exists;
- local-agent and server-LLM calls continue to use `T3MP3ST_API.baseUrl`;
- extracted `docs/index.html` scripts pass `node --check`.

No remaining blocker from my review.

jmagly:
Local verification passed on head `2f1fbd8` as `jmagly`:

```bash
npm ci
npm run typecheck
npm test
```

PR #79, opened 2026-07-12
feat: SOCKS5 egress proxy + harden local-model (llama.cpp/Ollama) path

SiNaPsEr0x:
> **Scope note:** this PR now carries **two independent bodies of work** in
> separate commits: (1) the original local-model hardening, and (2) a new
> **SOCKS5 egress proxy** for test traffic. They share no code; happy to split
> the proxy into its own PR if the maintainer prefers.

---

## Part B — SOCKS5 egress proxy for test traffic + egress-IP badge (new commit)

Lets an operator route **all probe/attack `fetch()` traffic** through a SOCKS5
proxy so tests don't leave from their own IP (rate-limits, WAF bans, attribution,
geo-testing), with a live egress-IP badge and a leak warning.

- **Single chokepoint**: `src/net/proxy.ts` installs a global undici dispatcher
  (`setGlobalDispatcher`) so every one of the ~32 scattered `fetch()` calls in the
  arsenal is proxied with no per-call edits. The dispatcher is an undici `Agent`
  with a custom `connect` that opens a SOCKS5 tunnel (via the `socks` client) and
  terminates TLS itself.
- **Loopback always bypassed** (`127.*` / `::1` / `localhost`): the local LLM
  (`:8070`) and the server's own same-origin calls never go through the proxy —
  routing them externally would break local mode.
- **Leak check** (`GET /api/net/ip`): fetches `ifconfig.co/json` twice — the exit
  IP (through the proxy when on) vs. the real IP (always-direct dispatcher) — and
  flags a leak when they match (no proxy, or proxy not effective).
- **UI**: top-right `IP:` badge next to Actions / War Room — 🟢 green = proxied,
  🔴 red = **LEAK** (real IP exposed), 🟠 amber = check failed; 30s poll. New
  **Settings → Egress Proxy** section (save/apply, disable, check-IP now).
- Config: `proxyUrl` setting + `getProxyUrl()/setProxyUrl()`; env `TEMPEST_PROXY_URL`
  wins. Endpoints `GET/POST /api/net/proxy` return a **credential-redacted** URL.

**Implementation notes (verified e2e):** Node's native `fetch` honours undici's
`setGlobalDispatcher` (same global symbol, `.2` on undici ≥ 8) — so a single
install of the `undici` package transparently proxies native fetch. But passing a
foreign `Agent` via the `{ dispatcher }` *option* throws `UND_ERR_INVALID_ARG`, so
the leak checker uses undici's own `fetch` to pass an explicit dispatcher. undici's
`httpSocket` upgrade path did not reliably terminate TLS on a SOCKS-provided socket,
so the connector does `tls.connect(...)` + `secureConnect` itself. Cross-checked
against the `fetch-socks` reference (same `Agent` + `connect` pattern; the loopback
bypass is our required addition it lacks). New deps: `undici`, `socks`.

**Part B verification:** local SOCKS5 server test — TLS-over-SOCKS fetch reaches
`example.com` through the tunnel (200 + real HTML), loopback correctly bypasses the
proxy (0 connections seen), and `ifconfig.co` correctly flags direct egress as a
leak. Endpoints smoke-tested (invalid URL → 400, creds redacted, enable/disable).
`tsc --noEmit` clean.

---

## Summary

Hardens the entire **local inference path** (llama.cpp / Ollama) so that slow
local reasoning models work reliably. Local reasoning models take tens of
seconds per call and separate their `<think>` trace from the final answer —
several code paths assumed fast cloud latency and a single content field, which
made local mode flaky or unusable. This PR fixes that end to end, plus a live
queue UI for sequential local calls.

Two intertwined bodies of work (kept in one commit because they touch the same
files):

### 1 — Local reasoning-model robustness + queue UX
- **Empty-response salvage**: `extractLocalContent()` 3-tier fallback
  (`content` → inline `<think>` unwrap → `reasoning_content`) so a model that
  exhausts its token budget mid-thought is no longer reported as an empty
  "possible refusal". Integrated for both the OpenAI-compatible and Ollama
  response shapes.
- **No premature abort on `/api/llm/local`**: listen to the *response* close
  event instead of the *request* close, so a slow local request is not aborted
  instantly as "Cancelled by operator".
- **LLM call queue popup**: real-time activity logging + persistent display for
  the single-slot local backend (serializes concurrent benchmark calls, live
  elapsed time, cancellation).
- Local-agent re-enable persistence + AI-backbone display-label fixes.

### 2 — Slow-local-model timeout audit (8 findings)
- **Scale every `safeLLMCall` site for local** via `llmTimeoutFor()`: the
  operator ReAct loop, SPECTRA `semanticAnalysis`, and self-improvement passed a
  bare 60s/25s timeout that aborted a slow reasoning model on every call (the
  ReAct loop then produced only error turns). Local connectivity probe raised to 90s.
- **Liveness probe no longer SIGKILLs a slow-but-alive local agent**:
  `pingLocalAgent` and the health check scale their timeout via env
  (`T3MP3ST_LOCAL_AGENT_PING_TIMEOUT_MS`), health floor 15s → 90s. Previously a
  healthy agent could be wrongly reported "connected but not live".
- **Structured local errors + smarter retries**: `LocalAdapter` throws
  `LLMApiError` on non-ok HTTP (so 401/403/404 are correctly non-retryable), and
  the retry loop treats a local timeout as permanent — re-firing an identical
  call on a single-slot backend just re-times-out and wastes ~6 min before
  falling through to a cloud model.
- **Preserve the real finish signal**: propagate `finish_reason` /
  `done_reason` instead of hardcoding `'stop'` (keeps `length` / `content_filter`).
- **Config**: the local provider gets a 120s timeout floor + `TEMPEST_LOCAL_TIMEOUT`
  override instead of inheriting the cloud-tuned default.

## New / changed config knobs
- `TEMPEST_LOCAL_TIMEOUT` — request timeout for the local provider (floors at 120s).
- `T3MP3ST_LOCAL_AGENT_PING_TIMEOUT_MS` — liveness-probe timeout for local CLI agents.
- `T3MP3ST_LOCAL_AGENT_HEALTH_TIMEOUT_MS` — health-check timeout (default 90s).
- `TEMPEST_PROXY_URL` — SOCKS5 egress proxy for test traffic, `socks5://[user:pass@]host:port` (Part B).

## Verification
- `tsc --noEmit` clean.
- The 5 pre-existing failing test files (2 static invariant + 3 bench
  `SyntaxError`) are unchanged vs `main` — not introduced by this PR (confirmed
  by stashing and re-running on a clean HEAD).
- 3 live local benchmarks pass against a llama.cpp reasoning model:
  KORP Terminal SQLi **71%**, Stack Overflow **63%**, Dynastic Cipher **62%**,
  ~40–50s each, no premature timeouts.

## Notes
- Files touched: `docs/index.html`, `src/llm/index.ts`, `src/server.ts`,
  `src/config/index.ts`, `src/agent/local-agents.ts`.
- No behavioural change for cloud providers; all timeout scaling is gated on
  local mode / the local provider.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -345,7 +345,31 @@
 
         .header-title { font-size: 18px; font-weight: 600; }
 
-        .header-actions { display: flex; gap: 10px; }
+        .header-actions { display: flex; gap: 10px; align-items: center; }
+
+        /* Egress IP badge — shows the IP the world sees for test traffic. Colour encodes
+           state: green = proxied (safe), red = real IP exposed (leak), grey = unknown. */
+        .egress-badge {
+            display: inline-flex; align-items: center; gap: 6px;
+            padding: 5px 10px; border-radius: 6px; cursor: pointer; user-select: none;
+            font-size: 11px; font-family: 'JetBrains Mono', ui-monospace, monospace;
+            border: 1px solid rgba(255,255,255,0.14); background: rgba(255,255,255,0.04);
+            transition: background .15s, border-color .15s;
+        }
+        .egress-badge:hover { background: rgba(255,255,255,0.08); }
+        .egress-badge .egress-dot { width: 8px; height: 8px; border-radius: 50%; background: #888; flex: 0 0 auto; }
+        .egress-badge .egress-label { color: #9aa; font-weight: 600; letter-spacing: .5px; }
+        .egress-badge .egress-ip { color: #cfd; font-weight: 700; }
+        .egress-loading .egress-dot { background: #888; animation: egressPulse 1s ease-in-out infinite; }
+        .egress-proxied { border-color: rgba(0,255,136,0.4); background: rgba(0,255,136,0.08); }
+        .egress-proxied .egress-dot { background: #00ff88; box-shadow: 0 0 6px rgba(0,255,136,0.7); }
+        .egress-proxied .egress-ip { color: #7CFFC4; }
+        .egress-leak { border-color: rgba(255,64,64,0.55); background: rgba(255,64,64,0.12); }
+        .egress-leak .egress-dot { background: #ff4040; box-shadow: 0 0 8px rgba(255,64,64,0.8); animation: egressPulse .8s ease-in-out infinite; }
+        .egress-leak .egress-label, .egress-leak .egress-ip { color: #ff8080; }
+        .egress-error { border-color: rgba(255,176,32,0.4); }
+        .egress-error .egress-dot { background: #ffb020; }
+        @keyframes egressPulse { 0%,100% { opacity: 1; } 50% { opacity: .35; } }
 
         .btn {
             padding: 8px 16px;
@@ -3524,6 +3548,11 @@
                 <button class="mobile-toggle" onclick="toggleSidebar()">☰</button>
                 <h1 class="header-title" id="pageTitle">War Room</h1>
                 <div class="header-actions">
+                    <span id="egressIpBadge" class="egress-badge egress-loading" onclick="refreshEgressIp(true)" title="Egress IP for test traffic — click to re-check">
+                        <span class="egress-dot"></span>
+                        <span class="egress-label">IP:</span>
+                        <span class="egress-ip" id="egressIpValue">…</span>
+                    </span>
                     <button class="btn btn-secondary" onclick="if(window.t3mpTour)t3mpTour.start()" title="Take the guided tour" aria-label="Guided tour" style="font-weight:700;">?</button>
                     <button class="btn btn-secondary" onclick="openQuickActions()">⌘ Actions</button>
                     <button class="btn btn-primary" onclick="navigateTo('warroom')">⚔️ War Room</button>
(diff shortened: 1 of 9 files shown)


jmagly (changes requested):
Thanks for the local-model and SOCKS proxy work. I found one behavioral regression that should be fixed before approval.

`src/net/proxy.ts` installs a process-global undici dispatcher and only bypasses loopback hosts (`localhost`, `127.*`, `::1`, etc.). The same PR adds/supports local-model `baseUrl` values for the operator's own "local/LAN model server" in `src/server.ts`, including per-request `/api/llm/local` calls. With the proxy enabled, a local model served from a private LAN address such as `http://192.168.1.50:8080/v1`, `http://10.0.0.5:11434`, or a private RFC1918 hostname is not bypassed; the LocalAdapter fetch is routed through the SOCKS proxy. In the common SSH/Tor/VPS SOCKS setups this either cannot reach the LAN host or sends local inference traffic to the egress proxy path, breaking local mode whenever the model is not on loopback.

The bypass should cover local/private destinations used for local model backends, or local LLM calls should use an explicit direct dispatcher instead of the global proxied dispatcher. At minimum, add coverage for RFC1918/link-local/ULA/private localhost-equivalent hostnames so enabling egress proxy for test traffic does not break LAN-hosted Ollama/llama.cpp/LM Studio.

Relevant lines from the PR head:
- `src/net/proxy.ts:85-91` only treats loopback as direct.
- `src/server.ts:297-300` documents support for the operator's own local/LAN model server.
- `src/server.ts:6817-6822` builds `/api/llm/local` with that client-supplied baseUrl, which then goes through LLMBackbone/LocalAdapter using global fetch.

Verification performed:
- `npm run typecheck`: pass
- `npm test`: pass (34 files / 386 tests)
- `npm run lint`: pass with warnings only, consistent with baseline
- Runtime smoke of `parseProxyUrl()` / `configureProxy()` status redaction: pass

No GitHub checks are currently reported for the PR branch, so this is from local review against `f9ac712`.

jmagly:
Flagged for HITL review

jmagly (approved):
Re-audited after the upstream rebase and integration repair. The local-model path now uses an explicit direct dispatcher whenever the attack-traffic SOCKS proxy is active, covering loopback, RFC1918/LAN addresses, and private DNS hostnames without weakening proxying for test traffic. Rebase conflicts and duplicate dependency/server/UI blocks were reconciled. Local gates pass: build, typecheck, 642 tests, 11 ops-preflight checks, lint (warnings only), npm audit (0 vulnerabilities), doctor (0 blockers), and diff check.

PR #83, opened 2026-07-13
fix(agents): detect local CLIs via well-known install dirs, not just PATH (macOS, #78)

### LINKED ISSUE #78: not detecting my Claude Code or Codex CLI.
I have both CLI's installed. i even checked the paths and everything is visible and normal however the Web UI refuses to detect them. i really don't wanna have to use an API key because i don't have one.

lyubomir-bozhinov:
## What & why

Refs #78 — the War Room "Connect Local Agents" panel reports an installed, logged-in **Claude Code / Codex** CLI as **Not installed** on macOS.

**Root cause (PATH scoping).** The detector resolves `claude`/`codex`/`hermes` against the **server process PATH** via `execFile`. A server started from **Finder, the desktop app, or any non-interactive shell** inherits launchd's minimal PATH — `/usr/bin:/bin:/usr/sbin:/sbin`. That omits every place these CLIs actually install:

- `~/.local/bin` — Claude Code's own native installer (symlinked into a versioned dir)
- Homebrew (`/opt/homebrew/bin`, `/usr/local/bin`), npm-global
- version-manager shims/bins — mise, nvm, fnm, asdf, volta, bun, …

So `execFile('claude', …)` throws `ENOENT`, the CLI reads `installed:false`, the checkboxes go dead, and it looks like a UI bug when the CLI works fine in the user's terminal.

## The fix

Add `resolveBin()` — **pure-fs, no shell** — that resolves a bin over `PATH` **plus** a curated set of well-known install dirs under the *real* agent home (`agentHome()`, so a HOME redirect can't hide them). Then:

- the detector (`detectOne`) and **both spawn sites** (`runLocalAgent`, `localAgentChat`) route through it — a CLI detected under a minimal PATH must also actually **spawn**, so "detected-but-unspawnable" can't happen;
- **fail-open**: an unresolved name is returned unchanged → the same `ENOENT` → `installed:false` as before, so true-absence detection is unchanged;
- `accessSync(X_OK)` **+ `statSync().isFile()`** so a directory named like the CLI can't be mistaken for the binary (matches `execvp`'s own skip-and-continue);
- PATH is scanned **before** the well-known dirs (shell precedence preserved);
- the keychain auth-probe gets a **2s timeout** so a locked/headless keychain (the same non-interactive launch context) can't hang detection.

No new dependencies. Replaces the old shell-based `command -v`/`which` lookup (which spawned `/bin/bash`) with an injection-free, hang-free scan.

## Scope

**POSIX / macOS.** `resolveBin` **short-circuits on `win32`** — Windows resolution (`where.exe`, `.cmd` shims) is the domain of the complementary open PR **#18** by @psigho, which this does not touch or duplicate. The two are additive: #18 covers Windows, this covers macOS/POSIX.

**Honesty on platforms:** developed and **confirmed on macOS (Apple Silicon)**. The resolver is the shared non-win32 path, so Linux launches exercise the same code and the generic dirs (`~/.local/bin`, `~/.nvm`, `~/.bun`, `/usr/local/bin`, …) apply there too — but I've only verified on macOS. Happy to extend the dir list if Linux users report gaps.

## Verification (macOS)

**Automated — RED→GREEN**, all green (`tsc --noEmit` clean, full suite **403 passing**, new file lints clean):

- `resolveBin` in isolation: well-known-dir hit under a minimal PATH; PATH-vs-well-known precedence; executable-file check (skips non-exec **and** a same-named directory); dynamic nvm/fnm per-version enumeration; symlink follow; `win32` short-circuit; qualified-path guard; fail-open.
- detector wiring: `detectLocalAgents()` reports `claude` `installed:true` with the resolved `path` + `version` when it lives only in `~/.local/bin` and PATH omits it.
- both rewired spawn sites (`runLocalAgent`, `localAgentChat`) launch the resolved binary (behavioral — a regression to the bare name would `ENOENT`).

**Manual QA — real Web UI, reproduced minimal PATH** (server booted with a PATH that runs node but omits `~/.local/bin`, so `claude` is invisible to the process — the exact #78 condition):

| | Claude Code row | summary | `/api/agents/local/detect` |
|---|---|---|---|
| **before** | `Not installed` · "binary not found on PATH" | `0/3 installed` | `installed:false` |
| **after** | `✓ Connected` · `v2.1.207` | `1/3 installed` | `installed:true`, `path:~/.local/bin/claude` |

(Codex/Hermes correctly stay `Not installed` — they genuinely aren't on this machine, so no false positives.)

## Notes

- Follows the CONTRIBUTING Review Standard (typecheck + full test suite pass locally).
- Minimal scope: 2 files, current `main` base, no file deletions, no unrelated changes.
- Security posture unchanged: no credential bytes read/logged; auth still detected by **presence** only.

Refs #78 — this fixes the **macOS/POSIX** variant. The original report (@HeavenDCS) turned out to be a **Windows Sandbox** environment, which is the domain of the complementary open PR #18, so this intentionally does **not** auto-close #78.

### DIFF
--- a/src/__tests__/local-agent-path-resolution.test.ts
+++ b/src/__tests__/local-agent-path-resolution.test.ts
@@ -1,0 +1,213 @@
+/**
+ * REGRESSION (issue #78 — macOS/POSIX): the local-agent detector resolved the CLI binary
+ * against the SERVER process PATH. When T3MP3ST is launched from Finder, the desktop app, or
+ * any non-interactive shell, launchd hands it a minimal PATH (/usr/bin:/bin:/usr/sbin:/sbin) —
+ * missing ~/.local/bin (Claude Code's native install), Homebrew, npm-global, and version-manager
+ * shims (mise/nvm/fnm). `execFile('claude', …)` then throws ENOENT and an installed, working CLI
+ * is reported `installed:false` (the Settings checkboxes go dead — reads like a UI bug).
+ *
+ * The fix is `resolveBin()`: pure-fs resolution over PATH + a curated set of well-known install
+ * dirs under the real agent home, with an executable-file check, fail-open to the bare name.
+ * These tests pin the resolver in isolation, the detector wiring, AND the two spawn call-sites
+ * (runLocalAgent / localAgentChat) the fix rewired — all through the public API.
+ *
+ * POSIX-only by design: `resolveBin` short-circuits on win32 (Windows resolution — where.exe /
+ * .cmd shims — is PR #18's domain, deliberately not duplicated here). The win32 case is asserted
+ * via a platform stub so the boundary is documented and locked.
+ */
+import { describe, it, expect, afterEach } from 'vitest';
+import { mkdtempSync, mkdirSync, writeFileSync, symlinkSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { resolveBin, detectLocalAgents, runLocalAgent, localAgentChat } from '../agent/local-agents.js';
+
+const FAKE_CLI = '#!/bin/sh\necho "1.2.3"\n';
+// For the chat path, which writes the prompt to the child's stdin: drain it before replying so the
+// parent's write()/end() completes cleanly (no EPIPE races), then emit a clean, non-error reply.
+const FAKE_CLI_STDIN = '#!/bin/sh\ncat >/dev/null 2>&1\necho "1.2.3"\n';
+const tmpDirs: string[] = [];
+const savedEnv = { PATH: process.env.PATH, AGENT_HOME: process.env.T3MP3ST_AGENT_HOME };
+
+/** A throwaway dir; tracked for afterEach cleanup. */
+function scratch(): string {
+  const d = mkdtempSync(join(tmpdir(), 't3mp3st-pathres-'));
+  tmpDirs.push(d);
+  return d;
+}
+
+/** Drop an executable (0o755) fake binary at `dir/name`; returns its absolute path. */
+function putExe(dir: string, name: string, content: string = FAKE_CLI): string {
+  mkdirSync(dir, { recursive: true });
+  const p = join(dir, name);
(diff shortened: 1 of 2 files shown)


jmagly (approved):
Reviewed PR 83 at `5916f19`.

The resolver preserves PATH precedence, skips relative/empty path entries, requires executable regular files, and uses the same resolution path for detection and both spawn sites. Windows is deliberately left to the existing bare-name behavior.

Verification:
- `npm run typecheck`: pass
- `npm test -- --run src/__tests__/local-agent-path-resolution.test.ts`: pass (repo script ran 35 files / 403 tests)
- `npm test`: pass (35 files / 403 tests)
- `npm run lint`: pass with warnings only, consistent with baseline

No blocking findings.

lyubomir-bozhinov:
Heads-up on scoping: I've changed `Closes #78` → `Refs #78` (PR body + commit `ed6b24f`) so this **doesn't auto-close** that issue.

After opening this, @HeavenDCS (who filed #78) noted their case was a **Windows Sandbox** environment — that's the domain of the complementary open PR #18, not this one. This PR fixes the **macOS/POSIX** PATH-scoping variant of the same detection bug (server launched with launchd's minimal PATH → `~/.local/bin`/Homebrew/version-manager CLIs invisible → `installed:false`), which reproduces and is verified independently on macOS. The two PRs are additive: #18 = Windows, this = macOS/POSIX. Happy to defer to the maintainer on how #78 should ultimately be closed.

jmagly (approved):
Re-audited current head `ed6b24f` after the force-push.

There is no code diff versus the previously approved `5916f19`; the update appears to be commit metadata/body wording around `Refs #78` rather than `Closes #78`.

Verification on `/tmp/t3mp3st-pr83-audit` at `ed6b24f`:
- `npm run typecheck`: pass
- `npm test -- --run src/__tests__/local-agent-path-resolution.test.ts`: pass (35 files / 403 tests)
- `npm test`: pass (35 files / 403 tests)
- `npm run lint`: pass with warnings only, consistent with baseline

No blocking findings.

Issue #90, opened 2026-07-14
Feature Proposal: Universal API Config & Local Agents Consolidation for T3MP3ST


synergy-ctf:
Hi Pliny / elder-plinius,

I hope you are doing well. First off, thank you for building T3MP3ST—it's an incredible platform for multi-agent offensive security and automated red-teaming!

I wanted to suggest a key design and usability improvement for the Settings page of the app. Currently, T3MP3ST has separate, redundant configuration sections for individual API keys (like OpenRouter, Venice, Anthropic, and OpenAI). 

To clean up the dashboard and make the orchestrations more flexible, I propose consolidating these into a unified, highly configurable "Universal API Config" panel, followed by a consolidated "Local Agents" panel. 

Here is a summary of the proposed interface structure:

1. Universal "API Config" Section:
- API Provider: A dropdown to select between 'OpenAI / OpenAI-compatible' and 'Claude (Anthropic Messages API)'.
- Consolidated Inputs: A unified set of inputs for Base URL, API Key, Model Selection (with a "Fetch list" button), and Max Context Tokens.
- Model Reasoning (Eino) Sub-section: A distinct, nested subsection to control reasoning settings (Model Reasoning toggle, Reasoning Effort selection, Wire profile, and "Allow chat page to override" toggle).

2. "Local Agents" Panel:
- Positioned directly below the main API config to list and manage CLI-authenticated agents on the local machine (Claude Code, Codex, OpenCode, Hermes, etc) with individual and bulk select "Connect" operations.

<img width="2543" height="1266" alt="Image" src="https://github.com/user-attachments/assets/3a1add64-2035-4947-a1e4-7b576aed2785" />

Best regards,

lyubomir-bozhinov:
Thanks @0xNoobody — this is a clean consolidation. I'd like to take it.

I recently landed the macOS local-agent detection fix (#83), so I've been deep in this exact Settings surface. Plan for one coherent PR:

**1. Universal API Config panel** — replaces the separate OpenRouter / Venice / Anthropic / OpenAI blocks with:
- Provider dropdown: `OpenAI / OpenAI-compatible` vs `Claude (Anthropic Messages API)`
- Unified inputs: Base URL, API Key, Model (+ a **Fetch models** button hitting the provider's list endpoint), Max Context Tokens
- a Model-Reasoning sub-section

**2. Consolidated Local Agents panel** — the Connect-Local-Agents card unified with the config area.

[...]

PR #101, opened 2026-07-16
feat(settings): Universal API Config panel with server-side model fetch (#90)

lyubomir-bozhinov:
## What & why

Implements the **Universal API Config** panel from #90 — one place in Settings to configure your active LLM provider instead of four separate key boxes.

**One coherent vertical slice** (server → UI): a provider dropdown (`OpenAI-compatible` / `Claude · Anthropic`), unified **Base URL / API Key / Model / Max Context Tokens**, and a live **Fetch models** button, sitting above the existing per-provider key sections.

## The slice

**Backend**
- `src/config/provider-models.ts` (new) — `listProviderModels` queries a provider's own model endpoint with the right wire shape (OpenAI-compatible `Bearer /models`, Anthropic `x-api-key /v1/models`, OpenRouter public list); `resolveModels` wraps it **fail-open**. 12 unit tests.
- `POST /api/models` (new route) — **server-side so the browser dodges provider CORS.** Always responds, failing open to the static `AVAILABLE_MODELS` list on a missing key, network error, or unknown provider. **Binds key+baseUrl by source** — a caller-supplied base URL must bring its own key, so the server never lends its configured key to a body-chosen URL.

**UI** (`docs/index.html`)
- The panel + handlers. **Backward-compatible by design**: it reads/writes the *same* `state.settings` keys the individual sections use (`${provider}Key`, `selectedModel`), so nothing migrates, nothing is lost, and the individual sections stay in sync.
- Model `<option>`s are built via the **DOM API** (`createElement`/`textContent`), never `innerHTML` — untrusted provider-returned model ids can't inject markup.

No new dependencies.

## Scope / honesty notes

- **Augment-not-replace.** I kept the individual per-provider key sections (removing them is a UX call for you — I asked in [the issue](https://github.com/elder-plinius/T3MP3ST/issues/90)). This panel is the primary surface; the old sections remain as the "advanced" fallback and stay in sync.
- **Base URL + context cap are stored** and used for model discovery / server + agent flows. In-browser mission routing still uses OpenRouter/Venice (matching the existing "not wired" hints on the Anthropic/OpenAI sections) — a small in-panel note says so rather than over-promising. Wiring a custom base URL into the browser chat path is a follow-up.
- **"Model Reasoning (Eino)"** from the proposal is a *separate* slice — it maps to nothing in the code today (only a hardcoded codex `model_reasoning_effort`), so it's deferred pending your answer on what it should map to.
- Touches `docs/index.html` in the same region **PR #18/#79** area; kept the markup change to a clean top-insertion and will rebase over #79 if it lands first.

## Verification (macOS)

- **12 unit tests** (RED→GREEN) for the model module — per-provider wire shape, base-URL normalization, fail-open.
- **Full suite 462 green** + 11 ops-preflight; typecheck clean; new files lint-clean.
- **Playwright QA — 17 checks** driving the real panel: renders, provider-switch updates the base-URL placeholder, **Save persists across reload**, **backward-compat** (the legacy `#openaiKey` input mirrors the same stored key), and both the **static** (real server, fail-open) and **live** (mocked) Fetch paths render correctly.
- **Two adversarial reviews** (security + correctness) run and their findings folded in before this PR: a DOM-XSS on model ids (→ DOM-built options), a server-key-to-body-URL disclosure footgun (→ key/baseUrl bound by source), and a **Critical** hang on unknown/`local-agent` providers (→ route always responds, fail-open) — all fixed and re-verified. One residual **Low** (unbounded provider-response read, same-principal self-DoS) is consistent with the codebase's existing uncapped user-URL fetches and noted for follow-up.

Refs #90

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -5515,6 +5515,49 @@
                 <!-- Settings -->
                 <div class="page" id="page-settings">
                     <div class="settings-section">
+                        <h2 class="settings-title">🎛️ Universal API Config</h2>
+                        <p style="color: var(--text-muted); font-size: 12px; margin-bottom: 12px;">Configure your LLM providers in one place — provider, key, base URL, model, and context cap — shared with the individual provider sections below (fully backward-compatible).</p>
+                        <div class="card">
+                            <div class="form-group">
+                                <label style="display:block; font-size:12px; color:#9fb0bd; margin-bottom:6px; font-weight:600;">API Provider</label>
+                                <select class="form-input" id="uacProvider" onchange="uacLoadProvider()">
+                                    <option value="openrouter">OpenRouter · OpenAI-compatible</option>
+                                    <option value="venice">Venice · OpenAI-compatible</option>
+                                    <option value="openai">OpenAI · OpenAI-compatible</option>
+                                    <option value="anthropic">Claude · Anthropic Messages API</option>
+                                </select>
+                            </div>
+                            <div class="form-group">
+                                <label style="display:block; font-size:12px; color:#9fb0bd; margin-bottom:6px; font-weight:600;">Base URL <span style="color:#6b7a85; font-weight:400;">— optional, blank uses the provider default</span></label>
+                                <input type="text" class="form-input" id="uacBaseUrl" placeholder="https://openrouter.ai/api/v1">
+                            </div>
+                            <div class="form-group">
+                                <label style="display:block; font-size:12px; color:#9fb0bd; margin-bottom:6px; font-weight:600;">API Key</label>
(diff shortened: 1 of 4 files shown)


jmagly:
Flagged for HITL review

jmagly (approved):
Maintainer audit complete on a085c38: provider/key source binding prevents credential forwarding to caller-chosen endpoints; origin/Host guards constrain the local SSRF surface; model IDs render through textContent; timeouts, fallback behavior, typecheck, 465 tests + 11 preflight checks, lint, dependency audit, doctor, and CI all pass.

PR #112, opened 2026-07-21
fix(ui): repair botched-merge syntax errors that break app boot (#111)

### LINKED ISSUE #111: APP doesn't initialize - unresponsive
After last update the APP is stuck in "Initializing..." status.

Also "Egress IP for test traffic" doesn't assume any IP.

The links in the APP don't work, the APP is stuck in the Homepage and the only thing we are able to do is scroll through it.  

In terminal, the only message continuously prompting is "[DATE/TIME] GET /api/selfimprove/ledger" .

lyubomir-bozhinov:
## What & why

Fixes #111 — the Web UI is stuck on **"Initializing…"**, all links/buttons dead, only `GET /api/selfimprove/ledger` looping in the terminal.

**Root cause:** `docs/index.html` is one big classic `<script>`. A series of bad auto-merges left **duplicated old/new copies of the same LLM-call code side by side**, producing JavaScript **syntax errors**. A syntax error means the browser never executes *any* of that script — so `navigateTo` and every handler are undefined, the health poll never runs (status stays "Initializing…"), and only a previously-scheduled poller keeps firing. The server itself is healthy (`/api/health` → `ok:true`).

Because a JS engine reports only the *first* syntax error per script, these were layered — each fix revealed the next. There were **eight**.

## The fixes (all in `docs/index.html`)

| Site | Scar | Fix |
|---|---|---|
| `_safeLLMCallOnce` | duplicate `response` decl + leftover pre-refactor `fetch` | one call, **backend-aware** endpoint/headers (OpenRouter **and Venice**, `backend.key`), cancellable `AbortController` kept |
| `safeLLMCall` | **entire function duplicated** (old truncated head + new loop referencing an undefined `backend`) | reunified: backend resolution + agent/server routing + model chain + fallback loop |
| `runLLMJudge` | two function bodies concatenated (one nested, never closed) | keep the complete `qItem`/cancellation variant |
| `runBenchmarks`, `autoApplyRecommendations` | old "ANY LLM backend" guard left unclosed before the new "reachable (key **or local mode**)" guard | drop dead old guard / restore missing `}` |
| benchmark auto-apply button, config-optimizer, self-improve, CTF call | duplicated option objects / double `catch` | collapse to the newer `_noFallback`/cancellable variant |

Net **−58 lines** (dead duplicated code removed). No behavior removed — the newer, local-mode-aware / cancellable variants are kept; Venice + OpenRouter + agent/server + model-fallback all preserved.

## Regression test

`src/__tests__/ui-inline-scripts-parse.test.ts` (new) compiles **every classic inline `<script>`** with `node:vm` — exactly how a browser parses it — with **zero new dependencies**. It fails on current `main` with `Identifier 'response' has already been declared` and passes after the fix, guarding the whole class of "the app won't boot because the script won't parse" regressions.

## Verification (macOS)

- **RED→GREEN**: the new test fails on `main`, passes after the fix.
- **Full suite 646 green**; typecheck clean.
- **Headless Chromium boot**: `#connectionText` now reads **"Connected"**, `typeof navigateTo === 'function'`, client-side navigation works, **0 uncaught JS errors** (was: stuck "Initializing…", `navigateTo` undefined).

Verified on macOS only. The LLM-call reconstructions are validated for parse + boot and preserve the pre-merge behavior of both merge halves; I couldn't exercise live provider calls without keys, so a maintainer spot-check of a real OpenRouter/Venice/local run is welcome.

Closes #111

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -11313,6 +11313,36 @@ Bypass all protections:
                 models = options._noFallback ? [primaryModel]
                     : (primaryModel === fallbackModel ? [primaryModel] : [primaryModel, fallbackModel]);
             }
+
+            let lastError = null;
+            for (const model of models) {
+                try {
+                    const content = await _safeLLMCallOnce(backend, prompt, model, options);
+                    // Detect empty/refused responses — model returned OK but no useful content
+                    if (!content || content.trim().length < 10) {
+                        console.warn(`[T3MP3ST] Model ${model} returned empty/refused response, trying next...`);
+                        addIntel('FALLBACK', `${model} → empty response, trying next model...`, 'warning');
+                        lastError = new Error(`Model ${model} returned empty response (possible refusal)`);
+                        continue;
+                    }
+                    if (model !== primaryModel) {
+                        console.log(`[T3MP3ST] Fallback to ${model} succeeded`);
+                        addIntel('FALLBACK', `Primary failed → using ${model}`, 'warning');
+                        fallbackCount++;
+                    }
+                    currentModelInUse = model;
+                    return content;
+                } catch (err) {
+                    console.warn(`[T3MP3ST] Model ${model} failed: ${err.message}`);
+                    addIntel('FALLBACK', `${model} error: ${err.message.substring(0, 80)}`, 'error');
+                    lastError = err;
+                    // Continue to fallback model
+                }
+            }
+            // All models failed
+            throw lastError || new Error('All models failed');
+        }
+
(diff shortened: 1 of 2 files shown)


beunwa:
It fix the problem for me (on linux)

lyubomir-bozhinov:
Thanks for testing, @beunwa — much appreciated. I could only verify on macOS, so a Linux confirmation is exactly the cross-platform datapoint this needed. 🙏

PR #113, opened 2026-07-22
test(ui): semantic defect gate for inline scripts (#111 hardening)

lyubomir-bozhinov:
## ✅ Now standalone — #112 has merged

Its dependency (**#112**, the fix for **#111**) merged on 2026-07-24, so I've **rebased this onto current `main`**. The diff is now just the one new commit — a single test file, **+49/−1**.

**Chain:** #111 (issue) → #112 (fix, merged) → **this PR** (the CI gate that stops #111 from recurring).

---

## What & why

#111 shipped **8 conflict-concatenation scars** into the 27,000-line SPA — and sailed straight through CI, because `lint`/`test` only ever look at `src/`. The product's primary surface has **no static gate at all**. #112's fix added a `vm.Script` *parse* test (catches outright syntax breakage). This PR adds the **semantic** layer parsing can't see — the shapes a bad auto-merge leaves behind that still parse but are always bugs.

## The gate

A new check in the existing test file runs **ESLint's high-precision "possible problem" rules** over each inline `<script>`:

`no-redeclare`, `no-unreachable`, `no-dupe-keys`, `no-dupe-args`, `no-dupe-else-if`, `no-duplicate-case`, `no-const-assign`, `no-import-assign`, `no-class-assign`.

Deliberately **minimal, additive, dependency-free:**
- **No new dependency** — `eslint` is already a devDependency; used via its `Linter` API.
- **No `eslint.config.js` or `ci.yml` change** — rides the existing `npm test`.
- **No flakiness** — pure static analysis, no browser/server. (A Playwright boot-smoke was considered and deliberately deferred: on a backendless CI its coverage largely overlaps this, while adding a browser dependency and timing flakiness.)

**Precision over recall** — each script is linted in isolation and the rule set is only "always-a-bug" checks, so it's false-positive-free on the real file. Two rules are excluded, reasons documented in-code:
- `no-undef` / `no-unused-vars` — the SPA relies on hundreds of cross-`<script>` and browser globals a per-block lint can't see (would flood).
- `no-func-assign` — the app legitimately monkey-patches `navigateTo` (wraps it to add CTF-range init).

## Validation

- **0 findings** on `main` (no false positives).
- **RED-proven**: the gate rejects the pre-fix tree (catches the duplicate `response`).
- Confirmed it catches each scar shape (dup-decl, unreachable-after-return, dup-key).
- `tsc --noEmit` clean; `npm run lint` adds 0 errors/warnings; **full suite 647 green**.

Relates to #111 / #112.

### DIFF
--- a/src/__tests__/ui-inline-scripts-parse.test.ts
+++ b/src/__tests__/ui-inline-scripts-parse.test.ts
@@ -14,6 +14,7 @@
 import { describe, it, expect } from 'vitest';
 import { readFileSync } from 'node:fs';
 import vm from 'node:vm';
+import { Linter } from 'eslint';
 
 const html = readFileSync(new URL('../../docs/index.html', import.meta.url), 'utf8');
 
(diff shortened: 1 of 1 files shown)


jmagly (approved):
Reviewed PR #113 at `837f3c7cd2cc5288eaa3be226fc0947789a6cff3`.

No blocking findings.

Verification:
- GitHub Actions T3MP3ST CI / test: success
- npm run typecheck: pass
- focused UI gate: 3/3 pass
- npm run lint: 0 errors, baseline warnings only

The semantic rules are narrowly scoped to high-confidence defects and complement the syntax gate without adding a dependency.

PR #114, opened 2026-07-23
fix(tools): run as non-root user and move binaries to /usr/local/bin

monkeygold:
## Summary
- Add non-root `operator` user so the Kali tools container no longer runs as root by default
- Move Go and Foundry binaries to `/usr/local/bin` (accessible to non-root) instead of `/root/go/bin` and `/root/.foundry/bin`
- Remove `/root` paths from `PATH` env

## Test plan
- [ ] `npm run tools:build` succeeds
- [ ] `npm run tools:check` passes (radare2/gdb/objdump/upx + python stack + pip-audit)
- [ ] Container runs as `operator` user: `docker run --rm cybench-tools:latest whoami` → `operator`
- [ ] Go tools accessible: `docker run --rm cybench-tools:latest nuclei -version`

### DIFF
--- a/ctf/docker-compose.yml
+++ b/ctf/docker-compose.yml
@@ -46,6 +46,11 @@ services:
     environment:
       - CTF_FLAG=T3MP3ST{bl1nd_sql1_t1m3_b4s3d}
     restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:80/"]
+      interval: 10s
+      timeout: 5s
+      retries: 3
     networks:
       - ctf-network
     labels:
@@ -66,6 +71,11 @@ services:
     depends_on:
       - webhook-receiver
     restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:80/"]
+      interval: 10s
+      timeout: 5s
+      retries: 3
     networks:
       - ctf-network
     labels:
@@ -85,6 +95,11 @@ services:
     depends_on:
       - metadata-mock
     restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:80/"]
+      interval: 10s
+      timeout: 5s
+      retries: 3
     networks:
       - ctf-network
       - ctf-internal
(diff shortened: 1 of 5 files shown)


jmagly (dismissed):
Reviewed PR #114 at `bb183694c502f8229d3c83b967b07b1c782ed0eb`.

Blocking finding:
- `npm run tools:build` fails at the changed `GOBIN=/usr/local/bin` layer. The Go install for ProjectDiscovery httpx reports that `/usr/local/bin/httpx` already exists and is not an object file. The earlier Python dependency layer installs an httpx console script there, so this path change creates a deterministic collision and the image cannot build.

Please stage or rename the ProjectDiscovery binary, or otherwise resolve the collision, then run the checked-off build, tools check, whoami, and representative Go-tool checks.

The compose healthcheck commit is also independent of the stated non-root scope; please split it or document and verify that scope explicitly.

Verification:
- docker compose config: pass, obsolete version warning only
- npm run tools:build: fail at Dockerfile Go-tool layer

monkeygold:
Addressed the blocking review in 2bd1493.

Changes:
- stage ProjectDiscovery Go binaries in /tmp/projectdiscovery-bin, then install them into /usr/local/bin so the Python httpx console script cannot make go install fail
- reuse Kali's existing operator group when creating the non-root operator user
- remove the unrelated compose healthcheck changes from this PR
- add npm run test:tools-dockerfile as a regression guard

Verification:
- npm run test:tools-dockerfile: PASS (7/7)
- npm run build --silent: PASS
- npm test: PASS (22 files, 292 tests)
- npm run lint: PASS (0 errors; 63 pre-existing warnings)
- docker compose config --quiet: PASS (existing obsolete-version warning only)
- Docker image build completed successfully, including the former failing Go-tool layer, and produced cybench-tools:latest with Config.User=operator and WorkingDir=/work
- presence smoke reported radare2, gdb, objdump, upx, Python imports, and pip-audit all OK

Please re-review the now-focused diff.

jmagly:
Maintainer follow-up on exact head `b9991bc`: the requested Go-binary staging fix and scope cleanup are complete. I added one narrow courtesy correction so the non-root image smoke uses a non-login shell (`bash -c`); Kali’s logout hook otherwise masked a successful smoke run with exit 1. Added regression coverage for that behavior. Full image build, non-root runtime checks, tools smoke, lint, typecheck, 292 tests, and exact-head CI all pass.

PR #115, opened 2026-07-23
feat(agent): Kimi Code CLI as keyless local-agent backbone

LexorCrypto:
## What

Adds **Kimi Code** (`kimi`, Moonshot AI's agentic CLI — Kimi K3 subscription login) as a fourth local-agent connector, alongside claude/codex/hermes. Same security posture: auth is detected **presence-only** (the login artifact `~/.kimi-code/credentials`, created by `kimi login` device-code OAuth — token bytes are never read), provider env keys are stripped before spawning, and the CLI's own login is the only credential used. No API keys, no new bills — the operator's existing subscription is the brain.

## Changes

- `src/agent/local-agents.ts`
  - `LocalAgentId` + `SPECS`: kimi entry — bin `kimi`, one-shot `kimi -p <prompt> --output-format text [-m alias]`.
  - `localAgentChat`: explicit kimi branch. Previously any non-claude/codex id fell through to the hermes `-z` argv, which kimi doesn't understand — every backbone call would have failed. kimi takes the prompt as the `-p` value; **stdout is reply-only** (thinking bullets and the session-resume trailer go to stderr — verified live), so no output filtering is needed.
  - `wellKnownBinDirs`: scan `~/.kimi-code/bin` (where the CLI installs; absent from launchd/desktop PATHs).
- `src/config/index.ts`
  - `getLLMConfig`: `case 'local-agent'` — agent id travels in `model`, no baseUrl, no API key. Without it, `TEMPEST_ORCHESTRATOR_PROVIDER=local-agent` (and friends) hit `default: throw Unknown provider`, so env-driven white-box/mission backbone selection could not use local agents at all.
  - Timeout floor for `local-agent`: **600s** with `T3MP3ST_LOCAL_AGENT_TIMEOUT_MS` as the operator knob. `LLMConfig.timeout` is passed as an explicit `timeoutMs` into `localAgentChat`, so a lower floor silently overrides the 600s dispatch default — this killed a real white-box run during verification (`kimi timed out after 120000ms` on a legitimate multi-minute planning prompt).
  - Model catalog: kimi entry, context 1,048,576 per https://openrouter.ai/moonshotai/kimi-k3.
- `src/llm/index.ts`: validate message lists kimi.
- `src/__tests__/local-agent-kimi-static.test.ts`: static invariants pinning all of the above (union, spec argv, chat-dispatch branch order, bin dir, catalog, config case, timeout floor).

## Verification (receipts)

Run labels: macOS arm64, kimi-code 0.29.0, CLI `default_model=kimi-code/k3` (effort high→max), harness = `npm run server` HTTP API. Agent id `kimi` is never passed as `-m` (the `opts.model !== id` guard) — calls ride the CLI's own default model alias.

- Live detection: `installed=true authed=true ready=true`, version 0.29.0 resolved from `~/.kimi-code/bin/kimi` under a minimal PATH.
- Live round-trips through the adapter: `pingLocalAgent` ok (~7s); `localAgentChat` with a strict-JSON prompt returned parseable output (orchestrator's stripFences + balancedSpans parser tolerates the CLI's leading `• `).
- End-to-end: two real white-box audits (`/api/whitebox/analyze`, local operator-owned repo, authorized) completed on the kimi backbone — 1 + 3 decomposition rounds, 32 worker/orchestrator round-trips, 0 refusals.
- Tests: 16/16 in the three local-agent suites incl. the new invariant file; `tsc` clean.
- Full `vitest run src`: **650/651**. The single failure (`local-agent-path-resolution.test.ts` — "reports Claude Code installed + versioned when claude lives in ~/.local/bin and PATH omits it") **reproduces identically on clean base `b192577`** in a scratch worktree without this patch: it's an environment-dependent test-isolation issue on machines that actually have `claude` installed at `~/.local/bin`, unrelated to this change. Left for a separate fix rather than growing this PR.

No headline numbers touched; no benchmark artifacts modified. No credentials, tokens, or target data read, printed, or committed — auth detection is presence-only by construction.

### DIFF
--- a/ctf/docker-compose.yml
+++ b/ctf/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       - CTF_FLAG=T3MP3ST{sql1_l0g1n_byp4ss_ez}
     restart: unless-stopped
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:80/health"]
+      test: ["CMD", "python", "-c", "import urllib.request;urllib.request.urlopen('http://localhost:80/health')"]
       interval: 10s
       timeout: 5s
       retries: 3
--- a/ctf/docker/web/sqli-basic/Dockerfile
+++ b/ctf/docker/web/sqli-basic/Dockerfile
@@ -9,9 +9,8 @@ WORKDIR /app
 # Install dependencies
 RUN pip install flask flask-cors
 
-# Create vulnerable application
+# Create vulnerable application (self-contained: templates are inline via render_template_string)
 COPY app.py /app/
-COPY templates/ /app/templates/
 
 # Set environment variables
 ENV FLASK_APP=app.py
--- a/docs/operations/audits/4496c4f-codex-audit.md
+++ b/docs/operations/audits/4496c4f-codex-audit.md
@@ -1,0 +1,14 @@
+# Codex audit — 4496c4f
+
+- **Commit**: `4496c4f` docs(status): defer 3 P2 kimi-adapter hardening findings from re-audit
+- **Auditor**: codex-cli 0.144.6, `gpt-5.6-sol`, reasoning effort high, `-s read-only`
+- **Date**: 2026-07-24
+
+## Verbatim verdict
+
+> [OK] Only `status.md` changed; all three P2 TODOs accurately mirror the audit records, with no secrets added.
+
+## Findings
+
+- [P1] none
+- [P2] none
--- a/docs/operations/audits/7182a53-codex-audit.md
+++ b/docs/operations/audits/7182a53-codex-audit.md
@@ -1,0 +1,21 @@
+# Codex audit — 7182a53
+
+- **Commit**: `7182a53` fix(ctf): make sqli-basic lab buildable and healthy
+- **Auditor**: codex-cli 0.144.6, `gpt-5.6-sol`, reasoning effort high, `-s read-only`
+- **Date**: 2026-07-24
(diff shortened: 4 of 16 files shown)


LexorCrypto:
Follow-up commit `9403c9a` added (driven by an independent read-only audit of the original patch):

- **`childEnv()` now strips the `KIMI_MODEL_*` family** — per the official env-var docs, `KIMI_MODEL_NAME` + `KIMI_MODEL_API_KEY` synthesize a temporary provider that would override the CLI's own OAuth login, quietly breaking the keyless posture. `TEMPEST_TARGET_HEADERS` / `TEMPEST_LOCAL_API_KEY` (t3mp3st's own secrets) are stripped too.
- **Auth detection targets the exact artifact** `~/.kimi-code/credentials/kimi-code.json` (an empty dir or one holding only `credentials/mcp/` is NOT a provider login), and `KIMI_CODE_HOME` is honored as a full data-root replacement.
- **Behavioral tests added** (`local-agent-kimi-behavior.test.ts`): real `oneShot` argv, `childEnv` stripping, `authState` under a relocated `KIMI_CODE_HOME`, and `getLLMConfig('local-agent','kimi')` resolution incl. the operator timeout override. The 600s value is documented as a *default* (env override honored for any positive value), and the catalog entry now notes the 1M window is tier-dependent.

Suites: 23/23 across the four local-agent test files; full `vitest run src` 657/658 (the same pre-existing environment-dependent failure noted in the PR description, reproduced on clean base).

LexorCrypto:
Closing per author decision — we're keeping the kimi local-agent backbone work in our fork for now (LexorCrypto/T3MP3ST). May re-propose upstream later. Thanks!

PR #117, opened 2026-07-24
fix(warroom): repair broken LLM-call sites so docs/index.html parses

hoptoad-off:
## What & why

The War Room UI (`docs/index.html`) ships with syntax errors in its main
`<script>` block from several bad-merge artifacts. Any one of them is a hard
`SyntaxError` that discards the entire ~20k-line block, so `navigateTo` is
never defined and sidebar navigation (Settings, Arsenal, …) does nothing — the
UI is stuck on the War Room page. `npm run verify-claims` doesn't cover the UI,
so this slipped through.

## Changes

- Merge the two split `safeLLMCall` definitions into one: keep the complete
  model-fallback loop and restore the `backend` resolution + keyless
  agent/server branch it referenced but no longer defined.
- Remove premature/dead fetch and options fragments in `_safeLLMCallOnce`,
  `runBenchmarks`, `runLLMJudge`, `generateAnalysis`,
  `autoApplyRecommendations`, `runSelfImprovementLoop`, the benchmark
  evaluator, and the CTF call — keeping the newer `_noFallback` /
  `llmTimeoutFor` variant in each.

Net: +17 / −85 (dead-code removal).

## Verification

- Every `<script>` block passes a syntax check; bracket balance is even.
- Headless browser: `navigateTo` defined (global fn count 93 → 515), all 14
  nav pages switch, Settings renders, no console errors.

## Notes

Semantics are preserved to the newer of each duplicated pair; no behavior was
invented. Suggest adding a lightweight UI syntax-check (per `<script>` block)
to CI so this can't regress.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -11290,29 +11290,6 @@ Bypass all protections:
         // Shared safe LLM call — resolves the backend, then routes the call. Keyless backends
         // (connected agent / server LLM) run single-shot through the server; key backends
         // (OpenRouter / Venice) call the provider directly and keep the model-fallback chain.
-        async function safeLLMCall(prompt, options = {}) {
-            const backend = resolveLLMBackend();
-            if (backend.kind === 'none') throw new Error(LLM_BACKEND_HINT);
-
-            if (backend.kind === 'agent' || backend.kind === 'server') {
-                if (typeof trackApiCall === 'function') trackApiCall();
-                const label = backend.kind === 'agent' ? ('agent:' + backend.agentId) : 'server-llm';
-                addIntel('API', `→ ${label}`, 'info');
-                const content = await _backendCall(backend, prompt, options);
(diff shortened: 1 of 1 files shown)


hoptoad-off:
Closing as a duplicate of #112.

I independently reproduced the same 8 botched-merge syntax scars in `docs/index.html` and arrived at the same reunified `safeLLMCall` / `_safeLLMCallOnce` fix (backend-aware endpoint for OpenRouter **and** Venice, AbortController kept) — but #112 by @lyubomir-bozhinov got there first, is already reviewed and independently confirmed, and is paired with the #113 CI parse/semantic gate. Deferring to it. I can add that I hit the exact same eight sites and the fix works on Linux, if that's a useful extra data point.

Separately I opened #118 for an unrelated backend bug that #112 doesn't touch: `getLLMConfig` throwing `Unknown provider: local-agent`, which aborts every keyless (connected-agent) mission.

PR #127, opened 2026-07-25
feat(llm): add Hugging Face provider + selectable local Claude Code models

2alf:
## Contribution Receipt
- Change: **Add Hugging Face provider (OpenAI-compatible Inference Providers router) and
  selectable local Claude Code models (active-agent pin + Opus/Sonnet/Haiku picker), UI + backend.**
- Scope class: local_lab
- Target authority: n/a (LLM backbone plumbing only; no target interaction)
- Network use: none (no live calls in tests; provider endpoints are config only)
- Run mode labels: n/a
- Model/harness labels: adds `huggingface` provider + `local-agent` model passthrough
- Commands run:
  - `npm run typecheck` -> pass
  - `npm test` -> pass except pre-existing Windows-only failures, identical set on clean upstream; new HF test passes
  - `npm run doctor` -> pass (optional-tool warnings only)
  - `npm run verify-claims` -> (no bench artifacts or headline numbers changed)
  - UI: all inline docs/index.html scripts pass `node --check`
- Artifacts: none
- Redaction: n/a (no secrets/evidence added)
- Claims changed: none
- Abstentions/refusals: none
- Residual risk: none beyond the existing local-agent execution path (unchanged trust model)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,9 +7,11 @@ VENICE_API_KEY=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 ANTHROPIC_API_KEY=sk-ant-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 OPENAI_API_KEY=sk-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 DEEPSEEK_API_KEY=sk-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+# Hugging Face — open models via the OpenAI-compatible Inference Providers router.
+# HF_TOKEN is canonical; HUGGINGFACE_API_KEY / HUGGINGFACE_TOKEN also work.
+HF_TOKEN=hf_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 GROQ_API_KEY=
 TOGETHER_API_KEY=
-HUGGINGFACE_TOKEN=
 REPLICATE_API_TOKEN=
 
 # ── Target authentication (both values are required) ──
--- a/docs/index.html
+++ b/docs/index.html
@@ -17159,14 +17159,17 @@ ACTION_INPUT: {"output":"Found confirmed SQLi on /login endpoint..."}`;
             }
             const apiKey = getApiKey();
             const savedProvider = localStorage.getItem('t3mp3st_general_provider');
-            const connectedAgent = (!apiKey && typeof window.t3mpPreferredAgent === 'function')
-                ? window.t3mpPreferredAgent()
-                : null;
-            const provider = savedProvider
-                || (connectedAgent ? 'local-agent' : undefined)
-                || (T3MP3ST_API?.connected && !T3MP3ST_API?.llmAvailable ? 'codex' : undefined);
+            // Centralized resolution: an explicit pin (e.g. Claude Code) outranks a stored key AND a
+            // remembered general-provider value, and the picked Claude model rides along as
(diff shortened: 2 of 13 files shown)


jmagly (changes requested):
Reviewed PR #127 at `d1ab2a7633edb8a5b200e7806cc936e32de0f911`.

Blocking findings:
1. The advertised keyless mission path is not self-contained on current main. This branch sends provider `local-agent`, but `ConfigManager.getLLMConfig` still lacks the local-agent case supplied separately by #118. Rebase after that dependency merges or include the fix and regression coverage here.
2. Model and pin routing is inconsistent. The main mission and safe LLM paths propagate the selected model and pin, but General and Admiral still select an agent only when no API key exists and send the bare agent id. A stored key can override the explicit pin there, and the selected Claude model is dropped. This contradicts the UI promise that the pin drives missions and analysis.

Please centralize backend resolution or update every dispatch surface and add tests for model argument passthrough, pin precedence, and General/Admiral routing.

Verification:
- npm run typecheck: pass
- focused provider/local-agent/UI tests: 44/44 pass
- npm test: 651/651 pass
- npm run lint: 0 errors, baseline warnings only
- No GitHub checks reported on this head

2alf:
Thanks for the review!
Both findings addressed in the latest commit:

1)
    Keyless path is now self-contained. Added the local-agent case to `ConfigManager.getLLMConfig`, so provider: 'local-agent' resolves keyless instead of hitting the default throw. 

2)
    Pin + model routing centralized. All three dispatch surfaces (mission, General, Admiral) now go through one shared resolver, `window.t3mpDispatchAgent`. 

Also tests have been added for all three: 
model-argument passthrough (adapter runs claude --model <picked>), 
pin-over-key precedence (evaluated from the shipped docs/index.html resolver), 
and a centralization invariant asserting all three surfaces route through the shared helper. 
Also updated the Op Admiral keyless routing static test to the new resolver.

Verification: typecheck pass; focused provider/local-agent/UI tests green; no net-new failures in npm test.

jmagly (changes requested):
Thanks for the substantial follow-up; the exact head passes typecheck, lint (warnings only), and the full 667-test suite. Two blockers remain:

1. This branch now conflicts with main after the narrower local-agent config fix in #118 was merged. Please rebase and reconcile the overlapping local-agent default deliberately (#118 uses `claude`; this branch uses `codex`) so the resulting behavior is explicit and tested.
2. The stale-pin regression test does not exercise the shipped `preferredAgent()`. In production, `preferredAgent()` returns a pinned agent when it is merely in `connectedIds`, even when it is absent from the live pool; `t3mpDispatchAgent()` then sees `pref === pin` and routes to that stale agent, including ahead of a valid stored API key. The test injects `preferredAgent: () => codex`, a state the real helper will not produce for a connected stale Claude pin. Please test the integrated resolver and ensure a non-live pin cannot displace a working keyed provider (or document and enforce a different intended degraded-mode policy).

jmagly (approved):
Thanks for completing the substantive review work. I used the enabled maintainer-edit permission to finish the remaining low-impact cleanup as a courtesy: merged current main, resolved the overlapping local-agent default in favor of main’s established `claude` behavior, and tightened the shared resolver so a connected-but-stale pin cannot displace a healthy keyed provider. I also added integrated live/stale pin regression coverage.

Exact-head verification at `c92e1c8`:
- mergeable cleanly against current main
- build/typecheck: pass
- lint: 0 errors (baseline warnings only)
- full suite: 64 files / 675 tests pass
- ops preflight and model-matrix/frontier self-tests: pass
- public-input/security preflight: low risk

Approved for squash merge.

PR #128, opened 2026-07-26
Wire garak into the Arsenal: structured parser + report-file capture

TengHu:
## What & why

garak was catalogued but `parserStatus: 'planned'` — an agent could run it, but its
output could only enter as the model's prose, which the live provenance gate
(`src/evidence/gate.ts`) refuses to mark verified. So the `ai_red_team` family had
**no verified, tool-backed evidence source**.

This wires garak end-to-end:

1. **`parseGarak`** (`src/arsenal/parsers.ts`) — turns garak's `report.jsonl` into
   `ToolFinding[]`, one finding per `(probe × detector)` that registered an attack
   success, with attack-success-rate → severity. It **summarises garak's own detector
   verdicts** (a parser summarises, it does not re-judge) and surfaces **only probe/
   detector names + counts — never prompt/response transcript text**.
2. **Report-file capture** (`src/arsenal/adapter-tools.ts` + `src/index.ts`) — garak
   writes its results to a `report.jsonl` **file**, not stdout. An additive
   `ArgTemplate.reportFile` hook + optional `mkReportPath`/`readToolReport` deps let the
   handler read that file back and use it as **both** the parse input **and** the evidence
   of record. **Zero behaviour change** for every stdout-based adapter.
3. `catalog.ts`: garak `parserStatus: 'planned' → 'structured'`.

**Result:** an `ai_red_team` mission can run garak against an authorized model and produce
findings that pass the provenance gate

## Scope honesty (please read)

- **"Verified" here = provenance-verified** ("garak really reported this"), **not**
  ground-truth-verified ("the model is really jailbroken"). garak's keyword detectors
  (`dan.DAN`, `mitigation.MitigationBypass`, `TriggerListDetector`) **over-report** — in
  local testing on a small model, a majority of `LatentJailbreak` "hits" were refusals
  (false positives). The finding `details` says so and recommends transcript verification.
  An **LLM-judge detector is the natural follow-up** (separate PR).

---

## Contribution Receipt

| Field | Value |
| --- | --- |
| **Change** | Wire garak into the Arsenal: `parseGarak` + report-file capture, so `ai_red_team` missions produce verified, tool-backed findings instead of unverifiable prose. |
| **Scope class** | `static_fixture` |
| **Target authority** | `not_applicable` (tests run against a committed static fixture; no live target in this change) |
| **Network use** | `none` |
| **Run mode labels** | `static_test`, `tool_backed` |
| **Model/harness labels** | `not_applicable` for the change's tests. The committed fixture is derived from a real garak 0.15.1 run against `meta-llama/llama-3.2-3b-instruct` via OpenRouter/litellm (author's own lab; not part of CI). |
| **Commands run** | `npm run typecheck` (`tsc --noEmit`) → **pass**. `npm test` (full `vitest run src`) → **651/652 pass**. The single failure — `local-agent-path-resolution.test.ts` (`codex` install detection, issue #78) — reproduces **identically on the unmodified baseline** (verified by stashing this change) and is environmental: the author's machine has the `codex` CLI installed, which the temp-home test does not expect. It is unrelated to this change and does not touch any file in it. `eslint` (changed files) → **clean**. |
| **Artifacts** | `src/__tests__/fixtures/garak.report.jsonl` (committed fixture). |
| **Redaction** | Parser surfaces only probe/detector names + counts, **never** transcript text (asserted by test). `readToolReport` consume-then-deletes the report file so probe transcripts do not linger on disk. Fixture contains no secrets. |
| **Claims changed** | `none` — no README/headline claim changed; garak is **not** added to any benchmark. |
| **Abstentions/refusals** | `not_applicable` (no model behaviour measured in this change). |
| **Residual risk** | (1) Findings reflect garak's over-reporting keyword detectors — transcript verification recommended; an LLM-judge detector is the natural follow-up (separate PR). (2) `promptfoo` remains `planned`. |

## Test evidence

- `parseGarak` field-mapping, aggregation, honesty-contract (empty/garbled → `[]`), and
  **no-transcript-leak** assertions.
- End-to-end: a parsed garak finding **passes `gateLiveFinding` with `provenance: 'tool'`**.

### DIFF
--- a/src/__tests__/adapter-tools.test.ts
+++ b/src/__tests__/adapter-tools.test.ts
@@ -10,6 +10,8 @@
  * All deps are injected fakes — no real binaries are spawned.
  */
 import { describe, it, expect } from 'vitest';
+import { access, stat } from 'node:fs/promises';
+import { dirname } from 'node:path';
 import {
   adapterToCustomTool,
   buildAdapterTools,
@@ -19,6 +21,7 @@ import {
   type AdapterToolDeps,
   type SubprocessResult,
 } from '../arsenal/adapter-tools.js';
+import { createPrivateReportWorkspace } from '../arsenal/report-workspace.js';
 import { TOOL_ADAPTERS } from '../arsenal/catalog.js';
 import type { ToolAdapter } from '../arsenal/catalog.js';
 import type { CustomTool, ToolContext } from '../types/index.js';
@@ -51,6 +54,18 @@ function makeDeps(overrides: Partial<AdapterToolDeps> = {}): AdapterToolDeps & {
 }
 
 const ctx = (parameters: Record<string, unknown>): ToolContext => ({ parameters });
+
+describe('report-file workspace', () => {
+  it('uses a private 0700 directory and cleanup removes the complete workspace', async () => {
+    const workspace = await createPrivateReportWorkspace('garak');
+    const dir = dirname(workspace.reportBase);
+    expect((await stat(dir)).mode & 0o777).toBe(0o700);
+    expect(dir).not.toBe('/tmp');
+
+    await workspace.cleanup();
+    await expect(access(dir)).rejects.toThrow();
+  });
+});
 
 describe('adapterToCustomTool — mint gate', () => {
   it('NEVER mints catalog_only / import_only adapters (metasploit, hydra, bloodhound → null)', () => {
--- a/src/__tests__/adapter-tools.test.ts
+++ b/src/__tests__/adapter-tools.test.ts
@@ -327,7 +327,7 @@ describe('invocation-honesty guard — every mintable adapter is classified, non
   // red-team harnesses, device-runtime tools, project-scaffolded RE, or a target-executing debugger).
   const OPERATOR_DRIVEN = new Set([
     'prowler', 'scoutsuite', 'cloudfox', 'pmapper', 'aws-cli', 'az-cli', 'gcloud-cli',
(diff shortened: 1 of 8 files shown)


jmagly (dismissed):
The exact head passes typecheck, lint (warnings only), and the full 652-test suite, and the structured parser itself looks sound. One security/privacy blocker remains: garak report JSONL can contain prompt/response transcripts, but it is created directly in the shared system temp directory and cleanup only happens after a successful read. A timeout, spawn error, non-zero garak exit, or read failure returns before deletion, leaving sensitive transcripts behind; direct `/tmp` creation also does not establish a private 0700 parent directory.

Please create reports inside a private per-run directory (for example `mkdtemp` with restrictive permissions) and remove the report/directory in a `finally` path covering success, non-zero exit, timeout, spawn error, and read failure. Add regression coverage for failure-path cleanup and the private temp location/permissions.

jmagly:
Maintainer courtesy cleanup completed at exact head `286d39e`. I replaced the shared `/tmp` report path with a private mode-0700 per-run workspace, moved deletion into an unconditional final cleanup path covering success/non-zero exit/timeout-spawn/read failures, and made garak fail closed when those workspace dependencies are unavailable. Added real filesystem permission/removal coverage and failure-path regression tests. Verification: 61 test files / 657 tests, ops preflight, model-matrix, refusal self-test, typecheck, lint, and exact-head CI all pass.

TengHu:
Thank you @jmagly

PR #140, opened 2026-08-01
fix(llm): populate usage for the Claude Code local-agent path

### LINKED ISSUE #139: Claude Code local-agent responses carry no `usage`: `AgentLoop`'s token budget check never fires
## Summary
On the connected-local-agent path (War Room → Settings → connect Claude Code), `LocalAgentAdapter.chat()` never populates `usage` on the `LLMResponse` it returns. `AgentLoop`'s per-task budget check depends on that field, so it silently never engages. The only thing left bounding a task is `maxIterations` (default 15), which counts turns, not spend. I pointed a mission at one authorized web target and it burned roughly **1.5 sessions of Claude Pro quota in about 30 minutes**, still only in kill-chain phase 2 of 7 (Weaponize).

## Repro
1. Connect Claude Code as the local agent in the War Room (no API key configured).
2. Launch a mission against an authorized target.
3. Watch real Claude Code usage climb while T3MP3ST's own token/budget accounting stays flat. The task-level `hitLimit` safety valve never engages, regardless of how long the mission runs.

## Root cause
`LocalAgentAdapter.chat()` never sets `usage`:

```ts
// src/llm/index.ts:1328-1343 (LocalAgentAdapter)
async chat(messages, options) {
  const content = (await localAgentChat(agentId, prompt, { model: agentModel, timeoutMs })).trim();
  return {
    content,
    model: `local-agent:${agentId}...`,
    finishReason: toolCalls?.length ? 'tool_calls' : 'stop',
    toolCalls,
    // no `usage` field: the CLI's stdout has no token counts to report
  };
}
```

`AgentLoop.run()`'s budget check depends on that field:

```ts
// src/agent/index.ts:214,336
tokensUsed += response.usage?.totalTokens || 0;   // always +0 for local-agent
if (tokensUsed >= this.options.maxTokens) {         // default 50_000
  hitLimit = true;
}
```

`tokensUsed` never leaves `0`, so the 50k-token circuit breaker is dead on this path. `maxIterations` is what's left, and it bounds turn count, not size. Each turn is also a stateless one-shot CLI spawn (`localAgentChat`, `src/agent/local-agents.ts:459`) that resends the entire growing transcript, no session reuse or caching, so cost per turn climbs roughly quadratically over a task. 

## Impact
Verified directly against the real code path (not just from reading the source), using a throwaway local sandbox target and `maxTokens` capped at an unrealistic 50: two real Claude Code turns ran (one issued a real tool call, one produced a 2,584-character debrief), `usage` was `undefined` both times, `tokensUsed` stayed `0`, `hitLimit` never engaged. A control run with the Mock adapter, which does return real `usage`, tripped `hitLimit` correctly on the very first turn under the same cap, so the check logic itself is fine. It's specifically the local-agent path where the number it needs is never populated.

## Suggested fix
Populate a token estimate in `LocalAgentAdapter.chat()`, e.g. a character-based heuristic (`Math.ceil((prompt.length + content.length) / 4)`) split into `promptTokens`/`completionTokens`, so `AgentLoop`'s existing budget check has something to compare against. Doesn't need to be exact, just needs to stop reading a permanently-zero counter.

N3thunt3r69:
Fixes #139.

`LocalAgentAdapter.chat()` never set `usage` on its `LLMResponse` for the Claude Code local-agent path, so `AgentLoop`'s per-task token budget check always compared against zero and never fired. `localAgentChat()` now requests `--output-format json` for the `claude` agent id only; `LocalAgentAdapter` parses the real envelope into `usage.promptTokens`/`completionTokens`, with a character-based estimate as a fallback if parsing fails. Codex and Hermes are untouched.

## Contribution Receipt
- Change: **Populate per-call usage for the Claude Code local-agent path (`--output-format json`) so `AgentLoop`'s token budget check can fire.**
- Scope class: local_lab
- Target authority: not_applicable (LLM backbone plumbing only; verification used a throwaway loopback fixture, not an authorized external target)
- Network use: loopback (verification target was a local throwaway HTTP server; the Claude Code CLI itself makes outbound calls to Anthropic under the operator's own account, inherent to exercising this adapter, not target network use)
- Run mode labels: local_agent, tool_backed
- Model/harness labels: model=claude-sonnet-5, provider=Anthropic (via Claude Code CLI, no API key), agent_runtime=Claude Code, harness=manual-review (ad hoc verification script, not a committed bench harness), tool_access=local_only
- Commands run:
  - `npm run typecheck` -> pass
  - `npm test` -> pass (675/675 vitest, 11/11 ops-preflight, 19/19 model-matrix)
  - `npm run doctor` -> pass (30/33, 3 warnings: semgrep/promptfoo missing, API health offline with the server not running — no blockers)
  - `npm run verify-claims` -> pass (27/27, no headline numbers changed)
- Artifacts: none committed. Verification used an ad hoc script (deleted after use) against a throwaway loopback HTTP fixture, plus two direct `claude -p --output-format json` calls confirming the envelope shape.
- Redaction: not_applicable (no secrets or evidence involved)
- Claims changed: none
- Abstentions/refusals: none
- Residual risk: Codex and Hermes local-agent paths are untouched and unverified (this fix targets `claude` only). The character-based fallback estimate (used only if the JSON envelope fails to parse) is a rough approximation, not exact. `agentFailureOutput`'s plain-text failure detection in `local-agents.ts` is not updated for JSON-wrapped error envelopes (pre-existing behavior, unchanged here).

### DIFF
--- a/src/__tests__/local-agent-tool-calling.test.ts
+++ b/src/__tests__/local-agent-tool-calling.test.ts
@@ -34,6 +34,7 @@ const TOOLS = [{
   parameters: { type: 'object' as const, properties: { target: { type: 'string' } }, required: ['target'] },
 }];
 const localBackbone = () => new LLMBackbone({ provider: 'local-agent', model: 'codex' } as never);
+const claudeBackbone = () => new LLMBackbone({ provider: 'local-agent', model: 'claude' } as never);
 const codexBackbone = () => new LLMBackbone({ provider: 'codex', model: 'codex-default' } as never);
 
 describe('parseTextToolCalls — happy path + drift tolerance', () => {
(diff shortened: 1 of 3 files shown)


jmagly:
Maintainer courtesy cleanup pushed as `11e3d4c`.

I kept the change narrow:
- preserve the parsed Claude `result` when the envelope has missing, empty, zero, malformed, or negative usage;
- accept real usage only when token fields are finite, non-negative, and non-zero in aggregate;
- otherwise invoke the existing character estimate instead of returning zero usage;
- add behavioral coverage for real cache-token aggregation, every fallback shape, non-JSON output, and tool-call parsing from the envelope result.

Verification on a synthetic merge into current `main` (`9020e5f`): typecheck passed; focused suite 33/33; full suite 748/748 Vitest plus 11/11 ops preflight and 19/19 model matrix; lint 0 errors (145 existing warnings on the merged tree). I’m waiting for hosted CI on the new exact head before merge.

jmagly (approved):
Reviewed at `11e3d4c884e05931e8728a427a6333888da86513`.

No blocking findings after maintainer-courtesy cleanup. Real Claude JSON usage is aggregated correctly; absent or invalid usage retains the parsed result and fails over to a non-zero estimate; non-Claude paths remain unchanged. Behavioral regression coverage exercises the reported budget-accounting boundary and JSON-wrapped tool calls.

Verification: hosted CI passed all gates; synthetic merge into current `main` passed typecheck, focused 33/33, full 748/748 Vitest plus ops/model checks, and lint with zero errors.

PR #142, opened 2026-08-01
feat(ingest): extract JS/TS arrow and function-expression definitions

### LINKED ISSUE #141: White-box ingest: JS/TS arrow functions and function expressions are never extracted
### Summary

White-box ingest extracts JS/TS **declarations only** (`function foo()`, `class C`, `method()`). Definitions bound to a *value* — arrow functions and function expressions — are silently not extracted, so their bodies never reach sink classification, priority scoring, or the context pack.

In modern Node/TS this is the dominant shape for exactly the code an operator cares about: Express handlers, Lambda entry points, exported utilities.

```js
const runCmd = (cmd) => exec(cmd);              // not extracted -> exec() sink unranked
module.exports.handler = async (e) => run(e);   // not extracted -> Lambda entry point invisible
const routes = { upload: (p) => open(p) };      // not extracted
```

This is a known limit of the multi-language ingest, documented in-code at the time it landed (#75) and tracked here as the follow-up.

### Where

`src/recon/ts-grammars.ts` — the `.js` / `.ts` / `.tsx` tree-sitter def-queries bind only:

```
(function_declaration ...) @def
(method_definition ...)    @def
(class_declaration ...)    @def
```

There is no pattern for `arrow_function` / `function_expression` bound to a name, so `parseFileMultiLang` returns nothing for those regions and the whole downstream pipeline (`classify` -> `prioritize` -> `findEntryPoints` -> `reachability` -> context pack) never sees them.

Reference — the caveat block in `src/recon/whitebox.ts`:

> JS/TS: only `function`/`method`/`class` declarations are captured. Arrow-function and function-expression definitions (`const f = () => …`) are not — idiomatic in modern TS, so their sinks may go unranked.

### Impact

- **Missed attack surface.** A file whose every function is `const f = (x) => …` currently ingests as zero blocks. Sinks in it (`exec`, `eval`, `fetch`, `open`) are never classified `attack_surface`.
- **Missed entry points.** `findEntryPoints` matches on name (`/^(handle|on|process|serve|…)/`). `module.exports.handler = async (e) => …` never produces a block, so it can never become an entry point — and nothing downstream of it is reachable.
- Silent, not noisy: the file parses fine, it just yields no blocks. Nothing in the output indicates the coverage hole.

### Proposed fix

Extend the `.js` / `.ts` / `.tsx` def-queries with the named-value function forms. No parser, pipeline, or `CodeBlock`-shape change — `nodeToCodeBlock` already handles any captured `@def`/`@name`/`@params` triple, and `kindOf` maps these to `function`.

Forms to cover:

| Form | Node |
|---|---|
| `const f = (a) => {}` / `const f = function (a) {}` | `variable_declarator` |
| `const f = a => …` (un-parenthesized single param) | `variable_declarator` |
| `obj.f = (a) => {}`, `module.exports.handler = …` | `assignment_expression` (name = the property, consistent with how `method_definition` is named) |
| `{ f: (a) => {} }` | `pair` |
| `class K { f = (a) => {} }` | `field_definition` (js) / `public_field_definition` (ts, tsx) |

Deliberately **out of scope** (would need a name-inference heuristic rather than a query, and are better as their own follow-up):

- HOC-wrapped definitions — `export const h = withAuth(async (req) => …)` — the value is a `call_expression`, so there is no direct name-to-function binding.
- Anonymous callbacks — `app.get('/x', (req, res) => …)` — no name to bind. (Related but separate: no decorator/route-registration entry-point elevation, also noted in the `whitebox.ts` caveat block.)

### Verification plan

RED test first: fixtures per language asserting each form is extracted with the right name/params/line, an end-to-end assertion that an arrow-bound sink ranks `attack_surface` and an arrow-bound `handler` is picked up as an entry point, plus negative assertions that non-function initializers (`const x = 42`), IIFEs, and destructured bindings are *not* matched, and that no existing declaration form double-extracts.

I hit this while working on #75 and have the fix. Happy to take it if no one else is on it.

lyubomir-bozhinov:
Closes #141.

White-box ingest extracted JS/TS **declarations only**. Every definition bound to a *value* — the dominant shape in modern Node/TS — produced no `CodeBlock`, so its body never reached sink classification, priority scoring, entry-point detection or the context pack. The file parsed fine and simply yielded nothing, so the hole was silent.

```js
const runCmd = (cmd) => exec(cmd);              // exec() sink went unranked
module.exports.handler = async (e) => run(e);   // Lambda entry point invisible
const routes = { upload: (p) => open(p) };      // not extracted
class C { #handler = (x) => eval(x); }          // not extracted
```

This is the follow-up to #75, where the limit was documented in-code when the multi-language ingest landed.

## The change

Query-only, in the `.js` / `.ts` / `.tsx` def-queries (`src/recon/ts-grammars.ts`). No parser, pipeline or `CodeBlock`-shape change: `nodeToCodeBlock` already handles any captured `@def`/`@name`/`@params` triple and `kindOf` maps these to `function`.

| Form | Node |
|---|---|
| `const f = (a) => {}` / `= function (a) {}` | `variable_declarator` |
| `const f = a => …` (un-parenthesized single param) | `variable_declarator` |
| `obj.f = …`, `module.exports.handler = …` | `assignment_expression` (member) |
| `onmessage = e => …` (worker/global handler) | `assignment_expression` (identifier) |
| `{ f: (a) => {} }` | `pair` |
| `class K { f = (a) => {} }`, `#f = …` | `field_definition` / `public_field_definition` |

Property-assigned and object-literal forms are named by the **property**, consistent with how `method_definition` is named — that is the name `buildCallGraph` and the entry-point heuristic match on. All three dialects share one pattern set; only the class-field node name differs (`field_definition property:` vs `public_field_definition name:`).

**Deliberately not matched**, because there is no name-to-function binding and inventing one is worse than honest absence:

- HOC-wrapped values — `const h = withAuth(async (req) => …)` (the value is a `call_expression`)
- anonymous callbacks — `app.get('/x', (req, res) => …)`

The `whitebox.ts` caveat block is updated in the same commit to say exactly this.

## Also fixed here (found by adversarial review of the change)

**Same-line block-id collision.** A block id is `path::name@line`, so two same-named definitions starting on the same line collide — and `buildCallGraph` keys on that id, so the two silently share one `callers`/`callees` entry even when only one carries a sink. Rare for declarations; routine once value-bound functions are extracted, since minified/bundled JS packs many `{ handler: … }` onto one line. The later block is now disambiguated by column. The first keeps the plain id, so **no existing id changes** and the Python path is untouched.

## Verification

RED first — the test commit fails on `main` (26 failures across all three dialects), passes after the query change.

- `npm test` — **708 passed**, 65 files (was 675 before this branch; +33 new). Green.
- `npm run test:coverage` — **100%** statements/branches/functions/lines on the three coverage-gated recon files, including the new branch in `ts-parse.ts`.
- `npm run typecheck` — clean. `eslint` on every changed file — 0 errors.
- `npm run doctor` — PASS (warnings are pre-existing optional-tooling/offline-API items).

The new suite pins both directions: each form extracted with the right name, params and span, **and** the excluded shapes (`const x = 42`, IIFE, destructured bindings, HOC-wrapped) asserted absent. The full-set equality assertion catches a missed form and an over-matching pattern in one check. An end-to-end test runs the real `ingestRepository` and asserts an arrow-bound sink ranks `attack_surface`, an arrow-bound `handler` becomes an entry point, and the call-graph edge between them resolves.

**Scale check** against this repo's own 1.24 MB / 21k-line SPA script (`docs/index.html`), old query vs new, same parse tree:

| | blocks | query match time |
|---|---|---|
| before | 560 | 32 ms |
| after | **679** | **28 ms** |

+119 real definitions recovered, no measurable match-time cost, and **zero** matches with an unbound `@name`/`@def` capture across that corpus — an unbound capture would throw and fail-open the whole file to `[]`, so this was checked explicitly rather than assumed.

Verified on **macOS (Apple Silicon)** only — that is the only platform I can test on. The change is a tree-sitter query string, so it carries no platform-specific behaviour, but I am not claiming Linux/Windows verification I did not do.

## Commits

| | |
|---|---|
| `f8988e0` | `test(ingest):` RED baseline — fails on `main` |
| `c8b0e20` | `feat(ingest):` the query change + caveat-block update |
| `7f34c47` | `fix(ingest):` `#private` fields + same-line id disambiguation (review findings) |

### DIFF
--- a/src/__tests__/ts-parse-arrow-fn.test.ts
+++ b/src/__tests__/ts-parse-arrow-fn.test.ts
@@ -1,0 +1,205 @@
+/**
+ * Regression guard for issue #141 — JS/TS arrow functions and function
+ * expressions were never extracted by white-box ingest.
+ *
+ * The def-queries in `ts-grammars.ts` bound only *declarations*
+ * (`function_declaration` / `method_definition` / `class_declaration`), so every
+ * definition bound to a value — `const f = (a) => …`, `module.exports.handler =
+ * async (e) => …`, `{ upload: (p) => … }`, `class K { f = (z) => … }` — produced
+ * no CodeBlock at all. Nothing downstream (classify → prioritize →
+ * findEntryPoints → reachability → context pack) ever saw those bodies, so their
+ * sinks went unranked and arrow-bound handlers could never become entry points.
+ * The file parsed fine; it just yielded nothing, so the hole was silent.
+ *
+ * These tests pin BOTH directions: every named-value form is extracted with the
+ * right name/params/span, and the shapes that carry no name-to-function binding
+ * (IIFEs, destructured bindings, HOC-wrapped definitions) stay unmatched rather
+ * than being guessed at.
+ */
+import { describe, it, expect, beforeAll } from 'vitest';
+import { mkdtempSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { tmpdir } from 'node:os';
+import { initGrammars, __resetGrammarsForTest } from '../recon/ts-grammars.js';
+import { parseFileMultiLang } from '../recon/ts-parse.js';
+import { ingestRepository, createMultiLangIngestConfig } from '../recon/code-ingest.js';
+
+beforeAll(async () => {
+  __resetGrammarsForTest(); // clear any singleton state leaked from another test file
+  await initGrammars();
+});
+
+/**
+ * One fixture, valid in all three dialects. Lines 2-13 must each yield exactly
+ * one block; lines 14-18 must yield none. Line numbers are asserted, so keep the
+ * leading newline and do not reorder.
+ */
+const FIXTURE = `
+const arrowBlock = (req, res) => { eval(req.body); };
+const arrowExpr = url => fetch(url);
+const arrowAsync = async (a, b) => { await run(a); };
+const funcExpr = function (cmd) { exec(cmd); };
+const namedFuncExpr = function inner(cmd) { exec(cmd); };
+obj.assigned = (u) => http.get(u);
+module.exports.handler = async (event) => { exec(event.cmd); };
+onmessage = e => eval(e.data);
+const routes = { upload: (p) => open(p), legacy: function (q) { open(q); } };
+class K { field = (z) => exec(z); m(a) { z(a); } }
+export const exported = (s) => exec(s);
+function plain(a) { a; }
+const notAFn = 42;
+const iife = (() => 1)();
+const [d1, d2] = [() => 1, () => 2];
+export const wrapped = mw(async (req) => exec(req));
+let noInit;
+`;
+
+/**
+ * Exactly the blocks the fixture must produce — nothing more. Asserting the full
+ * set (not just presence) is what catches BOTH a missed form and a pattern that
+ * double-extracts an existing declaration or matches a non-function initializer.
+ */
+const EXPECTED = [
+  'K',
+  'arrowAsync',
+  'arrowBlock',
+  'arrowExpr',
+  'assigned',
+  'exported',
+  'field',
+  'funcExpr',
+  'handler',
+  'legacy',
+  'm',
+  'namedFuncExpr',
+  'onmessage',
+  'plain',
+  'upload',
+];
+
+describe.each([
+  ['.js', 'a.js'],
+  ['.ts', 'a.ts'],
+  ['.tsx', 'a.tsx'],
+])('named-value function extraction (%s)', (ext, path) => {
+  const blocks = () => parseFileMultiLang(path, FIXTURE, ext);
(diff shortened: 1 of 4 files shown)


jmagly (approved):
Reviewed at `7f34c47b97b0605dfb7dad5edaff646659eccee2`.

No blocking findings. The query patterns cover the named-value forms without broadening into anonymous or wrapped callbacks, the same-line ID disambiguation preserves existing first-occurrence IDs, and the regression suite exercises extraction, exclusions, security ranking, entry-point elevation, and reachability.

Verification on a synthetic merge into current `main` (`a606adf`):
- `npm run typecheck`
- focused regression suite: 33/33
- full suite: 740/740 Vitest, 11/11 ops preflight, 19/19 model matrix
- hosted CI: pass

PR #151, opened 2026-08-04
feat(binary): detect memcpy/memmove with a non-constant length

### LINKED ISSUE #150: Binary/RE detector: no rule for memcpy/memmove with a non-constant (wire-controlled) length
### Summary

The Binary/RE decompiled-vuln detector (`scripts/binary-vuln-bench.mjs`) ships 6 rules — `gets`, `strcpy`, `sprintf`, format-string, `system`/`popen` injection, and `malloc`-multiply integer overflow — but has **no rule for `memcpy`/`memmove` with a non-constant length**. That is the classic unbounded-copy overflow, and it is a blind spot in exactly the pattern this project describes as its core target.

The reasoning layer already treats this as *the* textbook flaw:

- `src/orchestration/prompts.ts:60` — the worker "REFUSES the moment the code it is shown looks like a textbook flaw — e.g. **a wire-controlled length flowing into an unchecked memcpy**".
- `src/orchestration/prompts.ts:80` — "the smoking-gun line (e.g. **the unchecked memcpy of a wire length**)".
- `src/orchestration/index.ts:11` — example question: "Does this function validate the length parameter before the **memcpy**?".

So the LLM adjudication layer is built around this sink, but the fast static pre-filter can't surface it as a candidate at all. `strcpy` is covered; its `memcpy(dst, src, len)` sibling is not.

### The detection gap

`RULES` in `scripts/binary-vuln-bench.mjs:22-29` has no `memcpy`/`memmove` entry. A decompiled artifact like:

```c
void parse(char *pkt) {
  char buf[128];
  unsigned len = ntohl(*(unsigned *)pkt);
  memcpy(buf, pkt + 4, len);   // wire-controlled length into a fixed buffer
}
```

produces zero hits.

### Proposed rule

A **discriminating** rule — it must fire on a variable/wire-controlled length but stay silent when the length is a `sizeof(...)` expression or a numeric literal, otherwise it is a tautology (a bare `/memcpy/` match would flag every safe bounded copy too):

```js
{ id: 'B-MEMCPY',
  desc: 'memcpy()/memmove() with a non-constant length — candidate overflow from an unchecked/wire-controlled size',
  test: rx(/\b(?:memcpy|memmove)\s*\([^,]+,[^,]+,\s*(?!\d|sizeof\b)[^)\s][^)]*\)/) },
```

It inspects the **third argument**: fires on `len` / `hdr->len` / `*p` / `ntohl(hdr->len)` / `n - 4`, stays silent on `sizeof(dst)` / `16` / `0x40`. Validated against a 22-case dangerous/safe matrix — perfect discrimination.

### Scope (honest)

Same posture as the existing corpus note: this is **static pattern detection over decompiled output** — it surfaces a *candidate* sink, not a proven bug. It does not do taint analysis or prove the length is actually attacker-reachable (that is the reasoning layer's job). Constants are inlined in decompiled output, so a named-macro length isn't a concern by construction. Out of scope for this rule: additive/shift overflow precursors, `strcat`, exec-family injection — separate follow-ups.

### RED-first plan

1. Add a bad fixture (`memcpy` with a wire-decoded length, `expect: ["B-MEMCPY"]`) **and** a clean control with two safe `memcpy`s (`sizeof`-bounded + constant length) to `bench/binary-vulns/corpus.json`. The clean control is what forces a real regex — a naive `/memcpy/` rule false-positives on it and fails the discrimination gate.
2. Add direct assertions to `src/__tests__/binary-vuln-bench.test.ts`. Both go RED with no rule present.
3. Add the single `RULES` entry → GREEN.

I hit this reviewing the binary loadout and have the fix ready. Happy to take it if no one's on it.

lyubomir-bozhinov:
Closes #150.

Adds `B-MEMCPY` to the Binary/RE decompiled-vuln detector (`scripts/binary-vuln-bench.mjs`) — flags `memcpy`/`memmove` calls whose length argument is **non-constant** (a variable, struct field, deref, or decoded value), the classic unbounded-copy overflow. `strcpy` was covered; its `memcpy(dst, src, len)` sibling was the gap.

## Why this one

The reasoning layer already treats this sink as its textbook target:

- `src/orchestration/prompts.ts:60` — refuses on "a wire-controlled length flowing into an **unchecked memcpy**"
- `src/orchestration/prompts.ts:80` — "the smoking-gun line (e.g. the **unchecked memcpy of a wire length**)"
- `src/orchestration/index.ts:11` — example question: "Does this function validate the length parameter before the **memcpy**?"

So the LLM adjudication layer is built around this pattern, but the fast static pre-filter couldn't surface it as a candidate at all.

## The rule

```js
{ id: 'B-MEMCPY', desc: 'memcpy()/memmove() with a non-constant length — candidate overflow from an unchecked/wire-controlled size',
  test: rx(/\b(?:memcpy|memmove)\s*\((?:[^,()]|\([^()]*\))+,\s*(?:[^,()]|\([^()]*\))+,\s*(?!\d|sizeof\b)[^)\s][^)]*\)/) },
```

It inspects the **third argument**: fires on `len` / `hdr->len` / `*p` / `ntohl(hdr->len)` / `(size_t)len`, stays silent on `sizeof(dst)` / `16` / `0x40`. Two subtleties are load-bearing and documented in-code: the `[^)\s]` after the lookahead stops `\s*` from backtracking onto whitespace (which would leak every bounded form through), and args 1–2 match a single level of **balanced parens** (`(?:[^,()]|\([^()]*\))+`) so a pointer cast `(void *)dst` and a comma-bearing call `get(a,b)` are each consumed as one arg — neither mis-splits the list into a false positive on a bounded length. (Pointer casts on args 1–2 are the common decompiled shape; the first cut used a paren-free arg matcher and missed them — caught by a second adversarial review pass.)

## Honest scope

This is a **directional static pre-filter over decompiled output** — a *candidate* sink, matching the existing corpus's stated posture. It is not taint analysis and does not prove the length is attacker-reachable (that is the reasoning layer's job). Disclosed misses, enumerated in the rule comment and pinned as tests so the boundary is auditable:

- a computed length that starts with a digit (`8*count`) or with `sizeof` plus a term (`sizeof(x)+n`)
- a 1st/2nd arg with a **doubly-nested** paren (`wrap(get(a,b))`)
- the `__memcpy_chk` / `memcpy_s` / `wmemcpy` variants (the `_chk` form is usually the bounded, safe one)

Each disclosed limit is pinned by a test so the boundary is auditable. Mid-argument `/* ... */` comments aren't stripped (a comment obscuring a `sizeof`/literal would false-fire), but decompiled output does not emit inline comments.

Out of scope for this rule, as separate follow-ups: `strcat`, exec-family injection, additive/shift overflow precursors.

## Tests — RED-first, discriminating

- **RED** (`a0b7198`): a bad fixture (`memcpy` with a wire-decoded length) and a **clean control with two safe memcpys** (`sizeof`-bounded + a constant length) go into the corpus, plus direct assertions. The clean control is what forces a real regex — a bare `/memcpy/` rule false-positives on it and fails the zero-false-positive gate, so the test isn't a tautology. Both fail on `main`.
- **GREEN** (`bf7b7d6`): the single `RULES` entry.
- **Hardened** (`12022c5`): a first adversarial review found a false positive on nested-call args (`memcpy(dst, get_src(a,b), sizeof(dst))` — a bounded call that fired); expanded the test into an in-tree fires/silent/disclosed-limit matrix.
- **Cast fix** (`966a46f`): a second adversarial review (on the shipped state) found the paren-free tightening had over-corrected and stopped matching pointer casts — the common decompiled shape. Switched args 1–2 to balanced-paren matching, which fixes the original FP and the cast miss together. Validated against a 31-case fuzz oracle + a ReDoS sanity check.

Corpus goes 6+2 → 7+3; the `>=6` floors stay satisfied.

## Verification (macOS, Apple Silicon)

- `node scripts/binary-vuln-bench.mjs` self-test: **7/7 seeded sinks caught, 3/3 controls clean, 0 false-positives.**
- `npm test`: **755 vitest tests green** (+ ops-preflight 11/11, model-matrix, refusal-frontier 19/19).
- `npm run typecheck`: clean. `eslint` on the changed TS test file: clean. (The `.mjs`'s pre-existing node-globals `no-undef` lint is unchanged by this diff and outside CI's `eslint src/**/*.ts` scope.)

macOS-neutral: the rule is a regex over committed text fixtures — no platform-specific behavior. I can only verify on macOS, so I'm not claiming Linux/Windows runs I didn't do.

### DIFF
--- a/bench/binary-vulns/corpus.json
+++ b/bench/binary-vulns/corpus.json
@@ -44,6 +44,20 @@
       "content": "void fmt(char *input) {\n  char b[16];\n  sprintf(b, \"%s\", input);\n}\n"
     },
     {
+      "id": "memcpy-wire-length",
+      "desc": "memcpy with an attacker/wire-controlled length (unchecked) — buffer overflow",
+      "clean": false,
+      "expect": ["B-MEMCPY"],
+      "content": "void parse(char *pkt) {\n  char buf[128];\n  unsigned len = ntohl(*(unsigned *)pkt);\n  memcpy(buf, pkt + 4, len);\n}\n"
+    },
+    {
+      "id": "memcpy-bounded-control",
+      "desc": "Safe control — memcpy bounded by sizeof and by a constant length",
+      "clean": true,
+      "expect": [],
+      "content": "void copy(char *src) {\n  char dst[64];\n  memcpy(dst, src, sizeof(dst));\n  char hdr[4];\n  memcpy(hdr, src, 4);\n}\n"
+    },
+    {
       "id": "bounded-control",
       "desc": "Safe control — bounded strncpy/snprintf/fgets",
       "clean": true,
--- a/scripts/binary-vuln-bench.mjs
+++ b/scripts/binary-vuln-bench.mjs
@@ -26,6 +26,13 @@ export const RULES = [
   { id: 'B-FORMAT-STRING', desc: 'printf(var) — user-controlled format string', test: rx(/\bf?printf\s*\(\s*[a-zA-Z_]\w*\s*\)/) },
   { id: 'B-CMD-INJECTION', desc: 'system()/popen() on a variable — command injection', test: rx(/\b(system|popen)\s*\(\s*[a-zA-Z_]\w*/) },
   { id: 'B-INT-OVERFLOW', desc: 'malloc/calloc with multiplication — integer-overflow alloc size', test: rx(/\b(malloc|calloc|alloca|realloc)\s*\([^)]*\*/) },
+  // Inspects the length (3rd) arg: fires on a variable / field / deref / computed
+  // length (`len`, `hdr->len`, `ntohl(hdr->len)`), stays silent when it is a
+  // `sizeof(...)` expression or a numeric literal. The `[^)\s]` after the lookahead
+  // is load-bearing — without it `\s*` backtracks onto whitespace and every bounded
+  // form leaks through. Decompiled output inlines constants, so a named-macro length
+  // does not arise here.
+  { id: 'B-MEMCPY', desc: 'memcpy()/memmove() with a non-constant length — candidate overflow from an unchecked/wire-controlled size', test: rx(/\b(?:memcpy|memmove)\s*\([^,]+,[^,]+,\s*(?!\d|sizeof\b)[^)\s][^)]*\)/) },
 ];
 
 export function loadCorpus() {
--- a/scripts/binary-vuln-bench.mjs
+++ b/scripts/binary-vuln-bench.mjs
@@ -26,13 +26,19 @@ export const RULES = [
   { id: 'B-FORMAT-STRING', desc: 'printf(var) — user-controlled format string', test: rx(/\bf?printf\s*\(\s*[a-zA-Z_]\w*\s*\)/) },
   { id: 'B-CMD-INJECTION', desc: 'system()/popen() on a variable — command injection', test: rx(/\b(system|popen)\s*\(\s*[a-zA-Z_]\w*/) },
   { id: 'B-INT-OVERFLOW', desc: 'malloc/calloc with multiplication — integer-overflow alloc size', test: rx(/\b(malloc|calloc|alloca|realloc)\s*\([^)]*\*/) },
-  // Inspects the length (3rd) arg: fires on a variable / field / deref / computed
-  // length (`len`, `hdr->len`, `ntohl(hdr->len)`), stays silent when it is a
-  // `sizeof(...)` expression or a numeric literal. The `[^)\s]` after the lookahead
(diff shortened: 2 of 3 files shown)


jmagly (approved):
Reviewed exact head `966a46f` against current `main` (`afc9dad`). No blocking findings.

Verification:
- hosted `test` check: success
- `npm run typecheck`: pass
- `npm test`: 755/755 pass, plus ops/model/refusal gates
- `npm run verify-claims`: 27/27 pass
- focused corpus/rule review: the dangerous fixture, bounded controls, pointer-cast cases, nested-call false-positive case, and disclosed directional limits are all pinned

The change preserves the benchmark's candidate-detector posture rather than overstating taint proof. Approved for this exact head; re-review if the head or base changes.

PR #158, opened 2026-08-20
fix(obsidivm): actionable error when the range service is unreachable (#156)

### LINKED ISSUE #156: Benchmark showing error
All the benchmark are showing error, 
also running npm run obsidivm:evolve -- --hunter live --judge-model deepseek-v4-pro --accept-threshold 0.7 --max-gens 1 --target obsidivm --prune-after 3 shows FATAL: fetch failed

lyubomir-bozhinov:
Closes #156.

## What was happening

`npm run obsidivm:bench:*` / `obsidivm:evolve` died with a bare **`FATAL: fetch failed`** — no mention of the OBSIDIVM service, the URL, or how to start it. The `obsidivm:*` benchmarks are a thin HTTP client over the separate OBSIDIVM range service (`docs/OBSIDIVM.md`, default `http://127.0.0.1:4200`); `obsidivm-bench.mjs` fetches the target spec via the bridge before anything prints, so with the service down the raw `fetch` throws straight through. `--hunter stub` hits it too — "stub" only means "no LLM hunter"; the grading step still needs the service, which the help text didn't say.

(The deterministic benches — `binary`/`mobile`/`cve`/`cloud` — pass offline and were never affected; "all benchmarks error" in the report was the OBSIDIVM ones plus, for the `evolve --hunter live` command, a missing LLM key.)

## The fix

Wrap the `fetch` in the bridge's shared `call()` — the single choke point for **every** OBSIDIVM request (`getSpec`/`scoreText`/`createRun`/…), so one guard fixes them all — and classify the failure:

| Failure | Before | After |
|---|---|---|
| service down (ECONNREFUSED) | `fetch failed` | `OBSIDIVM service unreachable at <url> … Start it per docs/OBSIDIVM.md (python3 range.py, :4200), or set OBSIDIVM_URL` + `err.unreachable` |
| request timeout (AbortError) | `fetch failed` | `OBSIDIVM <m> <path> timed out after <n>ms (<url>)` |
| malformed `OBSIDIVM_URL` (ERR_INVALID_URL) | `fetch failed` | `Invalid OBSIDIVM base URL "<v>" — set OBSIDIVM_URL … to a full URL like http://127.0.0.1:4200` |

The HTTP-status path (`!res.ok`) and body parsing are untouched — the new `catch` wraps only the `fetch()` call, so it never swallows a status or JSON error. Also clarified the bench `--help`/usage that **all** hunter modes (stub included) need the OBSIDIVM service, and live/t3mp3st additionally a key.

## Tests

RED-first (`70a4d5d` fails on `main` with the bare `fetch failed`). Three hermetic cases in `src/__tests__/obsidivm-bridge.test.ts`, no mocking:
- **unreachable**: bind an ephemeral port, close it, connect → ECONNREFUSED → asserts the actionable message names the URL + escape hatches and sets `err.unreachable`.
- **timeout**: a TCP server that accepts but never responds + short `timeoutMs` → asserts `timed out after Nms` and **not** `unreachable`.
- **invalid URL**: asserts the config-error message points at `OBSIDIVM_URL`, not at starting a service.

Adds `scripts/obsidivm-bridge.d.mts` so the test types cleanly (matches the sibling `.d.mts` stubs for the other benches).

The invalid-URL branch, the timeout coverage, and the completed type stub all came from an adversarial self-review of the first cut, which had folded ERR_INVALID_URL into the "unreachable — is it running?" message.

## Verification (macOS, Apple Silicon)

- `npm test`: **758 vitest** green (+2 new branch tests) + ops-preflight 11/11, model-matrix, refusal-frontier 19/19.
- `npm run typecheck`: clean. `eslint` on the changed TS test: clean. Timeout test is stable across repeated runs (~300 ms, no hang). (The `.mjs` files carry a pre-existing node-globals `no-undef` lint, identical on `main` and outside CI's `eslint src/**/*.ts` scope.)
- Live: `OBSIDIVM_URL=not-a-valid-url npm run obsidivm:bench:stub` and a stopped service now print the actionable messages above instead of `fetch failed`.

macOS-only verification, but the change is provider/OS-neutral error handling.

### DIFF
--- a/scripts/obsidivm-bench.mjs
+++ b/scripts/obsidivm-bench.mjs
@@ -7,7 +7,9 @@
  *   2. Dispatch a t3mp3st hunter against the target's URL.
  *      - --hunter=live   : direct LLM (OpenRouter / Anthropic / OpenAI auto-detect)
  *      - --hunter=t3mp3st: drive t3mp3st's /api/general/auto (full platform)
- *      - --hunter=stub   : synthesized transcript (no LLM, for plumbing tests)
+ *      - --hunter=stub   : synthesized transcript (no LLM, for plumbing tests).
+ *                          NB: only the HUNTER is offline — the OBSIDIVM service
+ *                          is still required for the spec (step 1) and scoring (3).
  *   3. Submit the agent transcript to OBSIDIVM's /api/score/text.
  *   4. Persist the run via OBSIDIVM's /api/runs ledger + per-target sessions.
  *   5. Print a per-target table + suite aggregate (weighted % + grade).
@@ -105,9 +107,12 @@ function help() {
 Options:
   --target <id>           Target id (repeatable). Default: all 14 OBSIDIVM targets.
   --hunter <stub|live|t3mp3st>
-                          stub:    synthesized transcript (no LLM)
+                          stub:    synthesized transcript (no LLM; the OBSIDIVM
+                                   service is still required for spec + scoring)
                           live:    direct LLM call
                           t3mp3st: drive t3mp3st's /api/general/auto
+  All modes need the OBSIDIVM service (see --obsidivm / docs/OBSIDIVM.md); live
+  and t3mp3st additionally need an LLM key.
   --obsidivm <url>        OBSIDIVM base URL (default http://127.0.0.1:4200)
   --t3mp3st <url>         t3mp3st base URL (default http://127.0.0.1:3333)
   --model <name>          LLM model id (default claude-opus-4-7)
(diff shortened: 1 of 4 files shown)


jmagly (approved):
Reviewed exact head `13998844a45e474d5f86ff13cd77d7e3b8d5c062` against `main` at `6d4e017`.

No blocking findings. The change keeps HTTP-status/body parsing outside the network-error catch, distinguishes timeout and malformed configuration from an unreachable service, and makes the OBSIDIVM dependency explicit in help text. The three regressions exercise the intended failure branches without external network access.

Verification:
- `npm ci`: pass
- `npm run typecheck`: pass
- focused OBSIDIVM bridge tests: 3/3 pass
- hosted `test`: pass on this exact head
- local full suite: 756/758; the failures were the known PATH-sensitive Codex-presence assertion and the wall-clock parser budget while four PR suites ran concurrently, neither in this diff

Decision: approve this exact head. The review expires if the head, base, or hosted check changes.

